Issue #6, opened 2016-02-23
Run ZODB cache garbage collection when returning long lazy result sets


eprigorodov:
Background: if `ZCatalog.threshold` is set to non-None value then `ZCatalog.catalog_objects()` calls  `Connection.cacheGC()` after eash `threshold` calls. That helps client code to limit its memory usage during large reindexing operations.

Reading large result sets from `Catalog.search()` can produce problem similar to mass reindexing: storing too many objects in the ZODB cache, overcoming cache size limits and triggering MemoryOverflow.

Let's consider the use case: ZCatalog with >100K entries, a query that returns ~10K results and client code loop that calculates aggregate from values stored in one of metadata columns.

Such a loop instantiates 10K non-persistent "brains" records. Each "brain" record fills its data with the values pulled from metadata columns. Catalog metadata is stored as IOBTree. Reading one document record from IOBTree fetches the whole IOBucket object that contains metadata for up to 60 catalog entries. 100K big IOBTree has average height of two-three layers and fetching one leaf bucket from the tree bottom requires reading also several intermediate buckets.

With uniform distribution most of 10K result records would fetch metadata from different IOBuckets and reading them all will pull the whole metadata IOBTree into ZODB cache.

Proposal: call `Connection.cacheGC()` after instantiating each `ZCatalog.threshold` "brains" records. Use the same mechanism as is used for transaction savepoint during `ZCatalog.catalog_object()`, but do not make transaction savepoints, only run cache garbage collection.

eprigorodov:
Another implementation option: add new ZCatalog control, `read_threshold`, and use it in the way similar to `threshold`, but in read operations only.

eprigorodov:
[Unit tests](https://github.com/eprigorodov/Products.ZCatalog/blob/issue6-cacheGC-on-search/src/Products/ZCatalog/tests/test_zodb.py) that illustrate the problem: `test_unmaintained_search` fetches twice more IOBuckets than is allowed by the default cache size limit, `test_maintained_search` is broken and should be fixed by coming proposal implementation.

eprigorodov:
A sample implementation is [committed](https://github.com/eprigorodov/Products.ZCatalog/commit/c54b19ca5294de06f47a5826e4878fde016e1b4a) to the repository fork, in the branch issue6-cacheGC-on-search. It fixes `test_maintained_search` and ensures that after reading a long query cache size stays within known limit.

[...]

PR #7, opened 2016-02-26
fix for issue #6, Run ZODB cache garbage collection when returning large result sets

eprigorodov:
Proposed is branch that calls `Connection.cacheGC()` after each `ZCatalog.threshold` reads from a query result set within a single transaction.

That should prevent ZODB cache from being overfilled with metadata IOBuckets when reading result records from a long query in sufficiently large ZCatalog.

More details are given in the [issue itself](https://github.com/zopefoundation/Products.ZCatalog/issues/6), examples are provided in the [unit test `test_zodb.py`](https://github.com/eprigorodov/Products.ZCatalog/blob/issue6-cacheGC-on-search/src/Products/ZCatalog/tests/test_zodb.py).

### DIFF
--- a/src/Products/ZCatalog/Catalog.py
+++ b/src/Products/ZCatalog/Catalog.py
@@ -415,10 +415,16 @@ class Catalog(Persistent, Acquisition.Implicit, ExtensionClass.Base):
             record.append(attr)
         return tuple(record)
 
+    def _maintain_zodb_cache(self):
+        parent = aq_parent(self)
+        if hasattr(aq_base(parent), 'maintain_zodb_cache'):
+            parent.maintain_zodb_cache()
+
     def instantiate(self, record, score_data=None):
         """ internal method: create and initialise search result object.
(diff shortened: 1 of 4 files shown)


tseaver:
@eprigorodov Thanks for the patch!

I am +1 on the intent, but feel a little leery of the current implementation:  mostly, you had to patch so many places that I'm wishing for a refactoring of the "make a brain" bits.

eprigorodov:
Hi Tres, thank you for the review and for appraisal. 

I also felt uneasy patching three different places, but the intent was to be as least intrusive as possible. With your blessing I have added another [feature branch](https://github.com/eprigorodov/Products.ZCatalog/tree/issue11-refactor-brain-creation) that refactors all brain creation blocks into single method `Catalog.instantiate()`. Details of implementation are in the issue #11.

This pull request was rebased on top of that refactoring, so now its cache maintenance code is being called from a single point. Or if you would prefer then I can split all changes in two different pull requests.

Thank you.

PR #8, opened 2016-03-07
remove ZODB3 as direct dependency

tomgross:


setup.py
```diff
@@ -52,7 +52,6 @@
         'Record',
         'RestrictedPython',
         'zExceptions',
-        'ZODB3',
```
tseaver: Let's add `ZODB` and `BTrees` as an explicit dependency, since this product imports from them directly.

tomgross:
Travis says OK, after some version adjustments.

tseaver:
Thank you!

PR #9, opened 2016-03-30
Caching support for UnIndex

andbag:
Hi,

I have adopted the caching mechanism of DateRangeIndex in order to improve the performance of catalog queries calling repetitively indexes inherited from UnIndex (e.g. FieldIndex, KeywordIndex etc.). The performance enhancement becomes especially noticeable on big sites. For example, in plone are the indexes review_state, portal_type, allowedRolesAndUsers etc. within a request affected. 

Kind regards
Andreas

### DIFF
--- a/src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
+++ b/src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
@@ -108,12 +108,12 @@ class DateRangeIndex(UnIndex):
 
     security.declareProtected(view, 'getFloorValue')
     def getFloorValue(self):
-        """"""
+        """ """
         return self.floor_value
 
     security.declareProtected(view, 'getCeilingValue')
     def getCeilingValue(self):
-        """"""
+        """ """
         return self.ceiling_value
 
     manage_indexProperties = DTMLFile('manageDateRangeIndex', _dtmldir)
@@ -121,7 +121,7 @@ class DateRangeIndex(UnIndex):
     security.declareProtected(manage_zcatalog_indexes, 'manage_edit')
     def manage_edit(self, since_field, until_field, floor_value,
                     ceiling_value, REQUEST):
-        """"""
+        """ """
         self._edit(since_field, until_field, floor_value, ceiling_value)
         REQUEST['RESPONSE'].redirect('%s/manage_main'
                                      '?manage_tabs_message=Updated'
@@ -245,28 +245,22 @@ class DateRangeIndex(UnIndex):
             return None
 
         term = self._convertDateTime(record.keys[0])
-        REQUEST = aq_get(self, 'REQUEST', None)
-        if REQUEST is not None:
-            catalog = aq_parent(aq_parent(aq_inner(self)))
-            if catalog is not None:
-                key = self._cache_key(catalog)
-                cache = REQUEST.get(key, None)
-                tid = isinstance(term, int) and term / 10 or 'None'
+        cache = self._getCache()
+        if cache is not None:
+            tid = isinstance(term, int) and term / 10 or 'None'
+            if resultset is None:
+                cachekey = '_daterangeindex_%s_%s' % (iid, tid)
+            else:
+                cachekey = '_daterangeindex_inverse_%s_%s' % (iid, tid)
+
+            cached = cache.get(cachekey, None)
+            if cached is not None:
                 if resultset is None:
-                    cachekey = '_daterangeindex_%s_%s' % (iid, tid)
+                    return (cached,
+                            (self._since_field, self._until_field))
                 else:
-                    cachekey = '_daterangeindex_inverse_%s_%s' % (iid, tid)
-                if cache is None:
-                    cache = REQUEST[key] = RequestCache()
-                else:
-                    cached = cache.get(cachekey, None)
-                    if cached is not None:
-                        if resultset is None:
-                            return (cached,
-                                    (self._since_field, self._until_field))
-                        else:
-                            return (difference(resultset, cached),
-                                    (self._since_field, self._until_field))
+                    return (difference(resultset, cached),
+                            (self._since_field, self._until_field))
 
         if resultset is None:
             # Aggregate sets for each bucket separately, to avoid
@@ -276,7 +270,7 @@ class DateRangeIndex(UnIndex):
             until = multiunion(self._until.values(term))
 
             # Total result is bound by resultset
-            if REQUEST is None:
+            if cache is None:
                 until = intersection(resultset, until)
 
             since = multiunion(self._since.values(None, term))
@@ -285,7 +279,7 @@ class DateRangeIndex(UnIndex):
             # Merge from smallest to largest.
             result = multiunion([bounded, until_only, since_only,
                                  self._always])
-            if REQUEST is not None and catalog is not None:
+            if cache is not None:
                 cache[cachekey] = result
 
             return (result, (self._since_field, self._until_field))
@@ -297,7 +291,7 @@ class DateRangeIndex(UnIndex):
             since = multiunion(self._since.values(term + 1))
 
             result = multiunion([since, since_only, until_only, until])
-            if REQUEST is not None and catalog is not None:
+            if cache is not None:
                 cache[cachekey] = result
 
             return (difference(resultset, result),
--- a/src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
+++ b/src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
@@ -227,6 +227,16 @@ class DateRangeIndex(UnIndex):
                     else:
                         yield (key, len(value))
 
+    def _record_cache_key(self, record, resultset=None):
+        iid = self.id
+        term = self._convertDateTime(record.keys[0])
+        tid = isinstance(term, int) and term / 10 or 'None'
+        if resultset is None:
+            cachekey = '_daterangeindex_%s_%s' % (iid, tid)
+        else:
+            cachekey = '_daterangeindex_inverse_%s_%s' % (iid, tid)
+        return cachekey
+
     def _apply_index(self, request, resultset=None):
         """Apply the index to query parameters given in 'request', which
         should be a mapping object.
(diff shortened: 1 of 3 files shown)


src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
```diff
@@ -35,6 +35,7 @@
 
 from Products.PluginIndexes.common import safe_callable
 from Products.PluginIndexes.common.UnIndex import UnIndex
+from Products.PluginIndexes.common.UnIndex import RequestCache
```
tseaver: Unused import.

src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
```diff
@@ -252,33 +250,22 @@ def _apply_index(self, request, resultset=None):
         used.
         """
         iid = self.id
-        record = parseIndexRequest(request, iid, self.query_options)
+        record = parseIndexRequest(request, self.id, self.query_options)
```
tseaver: If you aren't going to use `iid` assigned above, delete the assignment too (it isn't used afterward).

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -36,6 +41,43 @@
...
+            self._misses +=1
+            return default
+        
+        self._hits += 1
+        return value
+
+    def __getitem__(self, key):
+        try:
+            value = super(RequestCache, self).__getitem__(key)
+        except KeyError, e:
```
tseaver: Use `except KeyError as e:`

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -36,6 +41,43 @@
...
+            self._misses +=1
+            raise KeyError, e
```
tseaver: Just `raise` would be better.

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -36,6 +41,43 @@
 _marker = []
 LOG = getLogger('Zope.UnIndex')
 
+class RequestCache(dict):
+    _hits = 0
+    _misses = 0
+    _sets = 0
```
tseaver: How do you see the stats being used?

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -304,6 +346,69 @@ def _apply_not(self, not_parm, resultset=None):
     def _convert(self, value, default=None):
         return value
 
+    def _catalog_cache_key(self, catalog):
+        cid = catalog.getId()
+        counter = getattr(aq_base(catalog), 'getCounter', None)
```
tseaver: I know it isn't new to this PR, but where are we expecting `getCounter` to come from?  It isn't a method of `ZCatalog` / `Catalog`.

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -304,6 +346,69 @@ def _apply_not(self, not_parm, resultset=None):
...
+        operator = record.get('operator', self.useOperator)
+        if operator != 'or':
+            # only operator 'or' is supported for caching
+            return None
+
+        not_parm = record.get('not', None)
+        if not_parm:
+
+            if isinstance(not_parm,list):
+                not_parm = tuple(not_parm)
```
tseaver: Not worth the trouble:  we are listifying immediately.

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -304,6 +346,69 @@ def _apply_not(self, not_parm, resultset=None):
...
+            elif not isinstance(not_parm,tuple):
```
tseaver: Use `if not isinstance(not_parm, (list, tuple)):`

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -304,6 +346,69 @@ def _apply_not(self, not_parm, resultset=None):
...
+                not_parm = (not_parm,)
+
+            not_parm = list(not_parm)
+            not_parm.sort()
+            not_parm = tuple(not_parm)
```
tseaver: Use `not_parm = sorted(not_parm)`

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -304,6 +346,69 @@ def _apply_not(self, not_parm, resultset=None):
...
+
+        usage_parm = record.get('usage', None)
+        if usage_parm:
+            params.append(('usage_parm', usage_parm))
+
+        kw = record.keys
+        if isinstance(kw,list):
+            kw = tuple(kw)
+        elif not isinstance(kw,tuple):
+            kw = (kw,)
```
tseaver: Ditto above coments for `not_parm`

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -424,7 +552,10 @@ def _apply_index(self, request, resultset=None):
                 for s in setlist:
                     # the result is bound by the resultset
                     r = intersection(r, s)
-
+                    # If intersection, we can't possibly get a smaller result
+                    if not r:
+                        break
+
```
tseaver: Don't we cache the intersected result?

src/Products/PluginIndexes/common/tests/test_UnIndex.py
```diff
@@ -70,3 +90,49 @@ def interesting(self):
...
+            res1 = idx._apply_index(record)
+
+            # Cache set?
+            self.assertEqual(cache._sets, expect) 
+
+            # Second Query
+            res2 = idx._apply_index(record)
+
+            # Cache hit?
+            self.assertEqual(cache._hits, expect)
```
tseaver: Shouldn't we assert misses / sets?

tseaver:
Thanks for the patch!  I'm not sure I've been able to grasp entirely how the cache interacts with `UnIndex._apply_index`, which is already quite complex. :(  I'd like another reviewer to look at it,  maybe @hannosch, @mauritsvanrees, @mgedmin, or @davisagli?

src/Products/PluginIndexes/common/UnIndex.py
hannosch: getCounter is only implemented in Plone's CatalogTool subclass at https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/CatalogTool.py#L367

It gets incremented whenever either catalog_object or uncatalog_object are called on the tool.

That logic should be moved to ZCatalog before it gets used here.

hannosch:
Just to refresh my memory, the way the date range index cache worked was:

- On the request object, it sets a RequestCache value for the key `catalog.getId()`, if getCounter is magically found it uses `'%s_%s' % (catalog.getId(), catalog.getCounter())`
- The date range index stores times rounded to minutes and the search terms get rounded to minutes to match this.
- The apply_index logic constructs a new cache id `'_daterangeindex_%s_%s' % (index.id, term / 10 or 'None')` and caches the resultset. Or tries to cache the inverse / difference of the resultset compared to an earlier query resultset if its smaller. 
- The getCounter and _increment_counter logic in Plone's CatalogTool only change when catalog_object or uncatalog_object are called on the ZCatalog. Not when methods are called on the Catalog or when methods are called directly on any index.

So this cache already can be wrong in a number of cases, when data changes during the same request. First do a query, then change some items in the catalog, then do a query again for the same term and the cache might not have been invalidated.

The extra 'divide by 10' to construct the cache key is also a bit weird. I think it was supposed to catch queries when you handle a request that takes a couple seconds and starts at second 59, going over to second 01. If something does repeated queries for "now" taking the exact minute would result in cache misses here. It would probably be better to use a relative difference here and consider cache entries if the difference between a cached value and the search term is below some threshold like 60 seconds instead.

In order to generalize this concept it might be better if the getCounter logic gets attached to each index to actually catch all changes to it, not just those done via the ZCatalog API. It might be best to implement such a change counter first in one PR and than let the cache generalization PR build on top of it.

mauritsvanrees:
I have trouble following what is happening in this code (not the pull request per se, but the catalog code in general), but the direction seems good. I agree with the detailed code quality remarks by Tres and the notes by Hanno.

andbag:
Thanks for the comments. I agree with Hanno's idea to implement the getCounter logic first in order to get rid of the dependency on plone. If you agree I create a PR for the getCounter logic and a separate PR for the generalized caching logic.

src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
```diff
@@ -288,7 +275,7 @@ def _apply_index(self, request, resultset=None):
             until = multiunion(self._until.values(term))
 
             # Total result is bound by resultset
-            if REQUEST is None:
+            if cache is None:
                 until = intersection(resultset, until)
```
jensens: This is not pull request related, but while reviewing this I stumbled over this:
We are in a block where it is guaranteed that ``resultset is None`` - why do we do ``intersection(None, until)``?

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -36,6 +41,43 @@
 _marker = []
 LOG = getLogger('Zope.UnIndex')
 
+class RequestCache(dict):
+    _hits = 0
+    _misses = 0
+    _sets = 0
+
+    def get(self, key, default=None):
+        value = super(RequestCache, self).get(key,_marker)
```
jensens: space is missing after `,`

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -36,6 +41,43 @@
...
+        
+        if value is _marker:
+            self._misses +=1
```
jensens: space is missing after operator (below in line 63 too)

src/Products/PluginIndexes/common/UnIndex.py
jensens: +1 for moving this logic here.

src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
hannosch: I noticed the same thing. That block can just be removed.

andbag:
PR #12 implements now the generalized caching logic based on the getCounter logic of master branch (incl. code quality corrections).

PR #10, opened 2016-04-07
Implementation of getCounter logic for UnIndex etc.

andbag:
I implemented the getCounter logic (incl. tests) in order to support the request based caching of UnIndex based indexes. Currently I need some feedback from you. :)

### DIFF
--- a/src/Products/PluginIndexes/BooleanIndex/BooleanIndex.py
+++ b/src/Products/PluginIndexes/BooleanIndex/BooleanIndex.py
@@ -62,6 +62,7 @@ class BooleanIndex(UnIndex):
         self._index_value = 1
         self._unindex = IIBTree()
         self._length = BTrees.Length.Length()
+        self._counter = BTrees.Length.Length()
 
     def histogram(self):
         """Return a mapping which provides a histogram of the number of
@@ -196,6 +197,7 @@ class BooleanIndex(UnIndex):
         """ Unindex the object with integer id 'documentId' and don't
         raise an exception if we fail
         """
+        self._increment_counter()
         unindexRecord = self._unindex.get(documentId, _marker)
         if unindexRecord is _marker:
             return None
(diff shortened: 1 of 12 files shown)


hannosch:
This looks really good to me. Would it be worthwhile to "document" the getCounter method and make it part of a new interface in PluginIndexes.interfaces? I wouldn't add it to the base IPluggableIndex, as there are other indices, like ZCTextIndex or elasticsearch/solr based index wrappers that might not want to implement this new method.

andbag:
A new interface is a good idea. However, I believe that a new interface makes sense additionally including the caching logic methods, i.e. getCache etc. Or is there another intended purpose for the getCounter method?

hannosch:
Right. A combined interface makes more sense. And I had trouble coming up with any name that describes getCounter on its own, something like IRequestCacheIndex is easier to spell out.

andbag:
Should be the new Interface 'IRequestCacheIndex' a part of this PR?

Which methods should be implemented for the new interface?

Fine-grained?

class IRequestCacheIndex(ILimitedResultIndex)
   def getCounter():
   def getCacheKey():
   def getCache(key):
   def getRecordKey(record, resultset=None):
   def getCachedResult(key):
   etc.

Or simpler only one method?

class IRequestCacheIndex(ILimitedResultIndex)
    def getRequestCachedResult(record, resultset=None):

hannosch:
I'd leave the interface out of this PR. Can't comment yet on the granularity question, as I haven't reviewed the other PR in detail.

I think this PR is fine as it is. Comments from anyone else?

tseaver:
The implementation looks fine to me (well, I wish there was better factoring between `UnIndex` and its derived classes, but that isn't the fault of this PR).  Can we have a docstring for `UnIndex.getCounter` explaining its purpose?

andbag:
Docstring added. :)

hannosch:
Merged as a squashed commit via github's UI.

PR #12, opened 2016-04-11
Request caching support for UnIndex 

andbag:
@hannosch 

I have now reimplemented the ideas of PR #9 based on the new getCounter logic. IMO, there are still open questions:
- Do we need an interface for the caching logic? Benefit?
- Should an index share a common dict per zcatalog for caching (as implemented) with the other indexes or should each index use its own dict?
- There are outstanding tests in order to check the functionality of the derived indexes. See example in [DateRangeIndex/tests.py](https://github.com/zopefoundation/Products.ZCatalog/compare/andbag-unindex-requestcache#diff-4b0783d021834f79034ac5ed35066ba7R273) (l. 273)

Last but not least, I need some feedback of you all. :)

### DIFF
--- a/src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
+++ b/src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
@@ -229,7 +229,6 @@ class DateRangeIndex(UnIndex):
                         yield (key, len(value))
 
     def _record_cache_key(self, record, resultset=None):
-        iid = self.id
         term = self._convertDateTime(record.keys[0])
 
         # why flatten to 10 minutes? 'term' is already flattened
@@ -237,11 +236,16 @@ class DateRangeIndex(UnIndex):
         # tid = isinstance(term, int) and term / 10 or 'None'
         tid = str(term)
 
+        # unique index identifier
+        iid = '_%s_%s_%s' % (self.__class__.__name__,
+                             self.id, self.getCounter())
+        # record identifier
         if resultset is None:
-            cachekey = '_daterangeindex_%s_%s' % (iid, tid)
+            rid = '_%s' % (tid, )
         else:
-            cachekey = '_daterangeindex_inverse_%s_%s' % (iid, tid)
-        return cachekey
+            rid = '_inverse_%s' % (tid, )
+
+        return (iid, rid)
 
     def _apply_index(self, request, resultset=None):
         """Apply the index to query parameters given in 'request', which
--- a/src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
+++ b/src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
@@ -18,10 +18,6 @@ from AccessControl.class_init import InitializeClass
 from AccessControl.Permissions import manage_zcatalog_indexes
 from AccessControl.Permissions import view
 from AccessControl.SecurityInfo import ClassSecurityInfo
-from Acquisition import aq_base
-from Acquisition import aq_get
-from Acquisition import aq_inner
-from Acquisition import aq_parent
 from App.Common import package_home
 from App.special_dtml import DTMLFile
 from BTrees.IIBTree import IITreeSet
@@ -40,12 +36,6 @@ from Products.PluginIndexes.interfaces import IDateRangeIndex
 
 _dtmldir = os.path.join(package_home(globals()), 'dtml')
 MAX32 = int(2 ** 31 - 1)
-
-
-class RequestCache(dict):
-
-    def __str__(self):
-        return "<RequestCache %s items>" % len(self)
 
 
 class DateRangeIndex(UnIndex):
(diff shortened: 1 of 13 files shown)


hannosch:
@andbag sorry, I've been busy. I'll try to get you a proper review this week(end).

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -324,6 +329,65 @@ def _apply_not(self, not_parm, resultset=None):
...
+                key = self._catalog_cache_key(catalog)
+                cache = REQUEST.get(key, None)
+                if cache is None:
+                    cache = REQUEST[key] = RequestCache()
+
+        return cache
+
+    def _catalog_cache_key(self, catalog):
+        # unique catalog identifier
+        cid = '_%s_%s' % (catalog.getId(), id(catalog))
```
hannosch: I'd add a constant prefix to this, like `_catalogcache_%s_%s` % (...)` - just to make it easier to spot what this is when looking at a request object. It would also lower the chance of key conflicts on the request object.

src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
```diff
@@ -238,12 +228,24 @@ def uniqueValues(self, name=None, withLengths=0):
...
-        counter = getattr(aq_base(catalog), 'getCounter', None)
-        if counter is not None:
-            return '%s_%s' % (cid, counter())
-        return cid
+    def _record_cache_key(self, record, resultset=None):
+        term = self._convertDateTime(record.keys[0])
+
+        # why flatten to 10 minutes? 'term' is already flattened
+        # to minutes by _convertDateTime.
+        # tid = isinstance(term, int) and term / 10 or 'None'
```
hannosch: I'd just remove this bit. I think I explained in a comment elsewhere why this was added, but its an edge-case. It's trying to let the cache kick in, if a requests takes longer than a couple seconds and crosses over the minute boundary. If the "same" query is issued, but with a slightly different time constraint at 15:01:59 and again 15:02:01 the cache key is different, as it based on the minute value. But there is a chance that the result should actually be different, so this isn't a totally safe assumption.

src/Products/PluginIndexes/common/UnIndex.py
```diff
@@ -324,6 +329,65 @@ def _apply_not(self, not_parm, resultset=None):
...
+            catalog = aq_parent(aq_parent(aq_inner(self)))
+            if catalog is not None:
+                key = self._catalog_cache_key(catalog)
+                cache = REQUEST.get(key, None)
+                if cache is None:
+                    cache = REQUEST[key] = RequestCache()
+
+        return cache
+
+    def _catalog_cache_key(self, catalog):
```
hannosch: Since this is only called from inside getRequestCache, I'd inline it into that function. Since the method on its own is only dependent on the catalog object but not the index, it otherwise feels like it's in the wrong place and should have been a method of the catalog. By simply inline it, that concern goes away.

src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
```diff
@@ -238,12 +228,24 @@ def uniqueValues(self, name=None, withLengths=0):
                     else:
                         yield (key, len(value))
 
-    def _cache_key(self, catalog):
-        cid = catalog.getId()
-        counter = getattr(aq_base(catalog), 'getCounter', None)
-        if counter is not None:
-            return '%s_%s' % (cid, counter())
-        return cid
+    def _record_cache_key(self, record, resultset=None):
```
hannosch: I think I'd rename this here and on the UnIndex to something like `getRequestCacheKey` and signal that it is a bit more of a public API.

Than we could define one interface that specifies both getRequestCache and getRequestCacheKey as its only two methods and UnIndex implement it. That interface would really just be for documentation purposes, but I think in Zope we still like to have interfaces for everything ;)

hannosch:
I put in some initial comments. As for your questions:
1. Interface: Zope kind of likes interfaces, so I still think it's the thing to do. I made a suggestion in the comment on how to do that.
2. Index vs. Catalog scope: I think the implementation is good as it is. Especially with the stats feel more appropriate on the catalog scope. Potentially one could in the future store some of the stats in the catalog query plan and persist them that way. That could help in deciding whether or not there is value in caching a query, or if it is typically just called once per request and just eats more memory.
3. Tests: Yeah, more tests would be really good. Especially for the intricacies that happen with the resultset logic.

I did a review of all but the real meat of this, which is the logic in the UnIndex. I'll need some time to wrap my head around that code again. It's hard to follow whether or not an intermediate value depends on just the query and index or also on the passed in resultset.

src/Products/PluginIndexes/common/UnIndex.py
andbag: Ok.

src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
andbag: IMO, if a request takes longer than a couple of seconds,  it should be normally an exception. The potential loss of a cached data and time performance is minimal (<100ms) in comparison with the complete duration of a long running request (>1s). The discretization of the index values underlies the same problematic for a long running request, isn't it? In addition, we have a problem with the tests, i.e. [DateRangeIndex/tests.py#L105](https://github.com/zopefoundation/Products.ZCatalog/blob/master/src/Products/PluginIndexes/DateRangeIndex/tests.py#L105). The passed times to the function _apply_index are only between the range [-1,15]. The round mechanism (10 min) collect most of them in the same class and the tests fail.

src/Products/PluginIndexes/DateRangeIndex/DateRangeIndex.py
andbag: :)

andbag:
Ok. If you agree, I will implement the "trivial" suggestions and look forward to hearing from you. BTW, it was already hard to follow the logic of UnIndex without caching and resultset feature. ;)

andbag:
The code is mostly revised and extended as discussed above:
  - Caching Interface added
  - Caching tests added

In particular, the caching tests for FieldIndex, KeywordIndex, UUIDIndex and DateIndex are subclasses of the default test class of each index. That implies all already existing search tests pass a caching test :). However, the subclass has the disadvantage that the tests are redundantly passed twice. Should I integrate the caching tests into the default tests?

hannosch:
@andbag I've finally found the time to review this. It actually all looked great. I only found one very small problem, I've fixed in 2d186a464686b1a7393e7cbfde3a0e5334dbf747. There was an edge-case where the cache could contain a single set for an `and` query. The cache retrieval logic worked, but would have iterated over all integers in the cached set one-by-one and done many intersect calls, instead of just a single intersect call.

The catalog tests could be nicer, and if you have time to clean them up further or avoid the duplication, please go ahead. But I didn't think that should block the merge.

PR #14, opened 2016-07-20
Consolidate and refactor request cache tests

andbag:
As suggested in PR #12, I have now cleaned up the tests for request caching support.


hannosch:
Excellent! :)

PR #15, opened 2016-07-29
CompositeIndex: Enhancement for Field- and KeywordIndexes 

andbag:
CompositeIndex is a plugin index for the ZCatalog. Indexes containing
     more than one attribute to index an object are called “composite
     index”. Such indexes should be created if you expect to run
     queries that will have multiple attributes in the search phrase
     and all attributes combined will give significantly less hits
     than the any of the attributes alone. The key of a composite
     index is called “composite key” and is composed of two or more
     attributes of an object.

     Catalog queries containing attributes managed by CompositeIndex
     are transparently catched and transformed seamlessly into a
     CompositeIndex query. In particular, large sites with a
     combination of additional indexes (FieldIndex, KeywordIndex) and
     lots of content (>100k) will profit. The expected performance
     enhancement for combined index queries is about a factor of >2-3.

     For example many catalog queries in plone are based on the combination of
     indexed attributes as follows: 'Language', 'review_state',
     'portal_type' and 'allowedRolesAndUsers'. Normally, the ZCatalog
     sequentially executes each corresponding atomic index and
     calculates intersection between each result. This strategy, in
     particular for large sites, decreases the performance of the
     catalog and simultaneously increases the volatility of ZODB’s
     object cache, because each index individually has a high number
     of hits whereas the the intersection between each index result
     has a low number of hits.

     CompositeIndex overcomes this difficulty because it already
     contains a pre-calculateted intersection by means of its
     composite keys. The loading of large sets and the following
     expensive computation of the intersection is therefore obsolete.

     IMPORTANT: CompositeIndex can only be used as an add-on not as 
     a replacement for field- and keyword indexes.

@hannosch
If a query of an atomic index contains a 'not' or 'and' operator CompositeIndex cannot catch the query and ZCatalog falls back to default intersection logic for this query.


hannosch:
@andbag thx for the PR! I'll try to have a look over the weekend :)

andbag:
thx :). Currently, the tests of CompositeIndex contain a benchmark between CompositIndex and "classical" search. The default number of indexed objects is 10000. You can increase the number (e.g. 100k) in [Line 165 of  testCompositeIndex.py](https://github.com/zopefoundation/Products.ZCatalog/blob/andbag-compositeindex/src/Products/PluginIndexes/CompositeIndex/tests/testCompositeIndex.py#L165) to get a first impression of the enhancement.

hannosch:
> Currently, the tests of CompositeIndex contain a benchmark between CompositIndex and "classical" search.

How long does that test take (in seconds?). If it's taking more than a second, it would be nice to put it onto a different test layer. That way you can run ``bin/test -u`` to only run the quick unit tests. I did something like that for the zodb cache tests in https://github.com/zopefoundation/Products.ZCatalog/blob/master/src/Products/ZCatalog/tests/test_zodb.py#L36. A test layer is just a class with setUp and tearDown class methods and gets referenced by the test class itself via a ``layer = TestLayer`` attribute.

src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py
```diff
@@ -0,0 +1,599 @@
...
+    meta_type = "CompositeIndex"
+
+    manage_options = (
+        {'label': 'Settings',
+         'action': 'manage_main',
+         'help': ('CompositeIndex', 'CompositeIndex_Settings.stx')},
+        {'label': 'Browse',
+         'action': 'manage_browse',
+         'help': ('CompositeIndex', 'CompositeIndex_Settings.stx')},
+    )
```
hannosch: Can you remove the help part of the manage_options? HelpSys got removed with Zope 4. If there's documentation that should be available, it could go into the add or manageCompositeIndex pages, or just live in the README.

src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py
```diff
@@ -0,0 +1,599 @@
...
+
+    query_options = ("query", "operator")
+
+    def __init__(self, id, ignore_ex=None, call_methods=None,
+                 extra=None, caller=None):
+        """Create an composite index
+
+        UnIndexes are indexes that contain two index components, the
+        forward index (like plain index objects) and an inverted
```
hannosch: This looks like copy/paste. I'd either remove it or update it ;)

src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py
```diff
@@ -0,0 +1,599 @@
...
+        self.clear()
+
+    def _apply_index(self, request, resultset=None):
+        """ Apply the index to query parameters given in the request arg. """
+        record = parseIndexRequest(request, self.id, self.query_options)
+
+        if record.keys is None:
+            return None
+
+        return UnIndex._apply_index(self, request, resultset=resultset)
```
hannosch: _apply_index looks like it doesn't actually do anything different, compared to UnIndex._apply_index. Did you mean to add anything here? Otherwise you could just remove it here and directly rely on the inherited method.

src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py
```diff
@@ -0,0 +1,599 @@
...
+            # we have an existing entry for this document, and we need
+            # to figure out if any of the keywords have actually changed
+            if type(oldKeywords) is not OOSet:
+                oldKeywords = OOSet(oldKeywords)
+            newKeywords = OOSet(newKeywords)
+            fdiff = difference(oldKeywords, newKeywords)
+            rdiff = difference(newKeywords, oldKeywords)
+            if fdiff or rdiff:
+                # if we've got forward or reverse changes
+                self._unindex[documentId] = list(newKeywords)
```
hannosch: KeywordIndex._index_object has additional logic for deletes here, maybe those should also go here:
```
# if we've got forward or reverse changes
if newKeywords:
    self._unindex[documentId] = list(newKeywords)
else:
    del self._unindex[documentId]
if fdiff:
...
```

src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py
```diff
@@ -0,0 +1,599 @@
...
+
+        if keywords is not None:
+            for kw in keywords:
+                self.removeForwardIndexEntry(kw, documentId)
+
+    def unindex_object(self, documentId):
+        """ carefully unindex the object with integer id 'documentId'"""
+
+        keywords = self._unindex.get(documentId, None)
+        self.unindex_objectKeywords(documentId, keywords)
```
hannosch: KeywordIndex has the new addition of:
```
if keywords is not None:
    self._increment_counter()
```

src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py
```diff
@@ -0,0 +1,599 @@
...
+        try:
+            del self._unindex[documentId]
+        except KeyError:
+            LOG.debug('%s: Attempt to unindex nonexistent '
+                      'document with id %s' %
+                      (self.__class__.__name__, documentId),
+                      error=sys.exc_info())
+
```
hannosch: The above three functions looks almost identical to the KeywordIndex. Would it be possible to inherit from KeywordIndex instead of UnIndex? That might cut down on the code duplication of those methods.

src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py
```diff
@@ -0,0 +1,599 @@
...
+            attr = component.attributes[-1]
+            try:
+                datum = getattr(obj, attr)
+                if safe_callable(datum):
+                    datum = datum()
+            except (AttributeError, TypeError):
+                datum = _marker
+            if isinstance(datum, list):
+                datum = tuple(datum)
+            return (datum,)
```
hannosch: Could you shorten this, by using UnIndex' _get_object_datum method? For example:
```
if component.meta_type == 'FieldIndex':
    attr = component.attributes[-1]
    datum = self._get_object_datum(obj, attr)
    if isinstance(datum, list):
        datum = tuple(datum
    return (datum,)
```

src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py
```diff
@@ -0,0 +1,599 @@
...
+                        # unique
+                        newKeywords = set(newKeywords)
+                    except TypeError:
+                        # Not a sequence
+                        datum.append(newKeywords)
+                    else:
+                        datum.extend(newKeywords)
+
+            datum.sort()
+            return tuple(datum)
```
hannosch: Assuming you'd inherit from KeywordIndex, this could be shortened as well:
```
elif component.meta_type == 'KeywordIndex':
    for attr in component.attributes:
        datum = []
        attr_datum = KeywordIndex._get_object_keywords(self, obj, attr)
        if attr_datum:
            if isinstance(attr, tuple):
                datum.append(attr_datum)
            else:
                datum.extend(attr_datum)
        datum.sort()
        return tuple(datum)
```

hannosch: And maybe at the same time, port the use of `set` to KeywordIndex, where it still uses a dict to create a set (``unique = {}``).

src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py
```diff
@@ -0,0 +1,599 @@
...
+        del self._components[c_id]
+
+        self.clear()
+
+    def saveComponents(self, components):
+        # Change the component object specified by 'c_id'.
+        for c in components:
+            self.delComponent(c.old_id)
+            self.addComponent(c.id, c.meta_type, c.attributes)
+
```
hannosch: Should this also do a ``self.clear()``, like add and del?

src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py
```diff
@@ -0,0 +1,599 @@
...
+        if len(c_meta_type) == 0:
+            raise RuntimeError('No component type set')
+
+        self.addComponent(c_id, c_meta_type, c_attributes)
+
+        if RESPONSE:
+            RESPONSE.redirect(URL1 + '/manage_main?'
+                              'manage_tabs_message=Component%20added')
+
+    def manage_delComponents(self, del_ids=[], URL1=None,
```
hannosch: I prefer using an empty immutable tuple, instead of a mutable list in function signatures. It's too easy to get bitten by `del_ids` being a mutable list defined at class scope later on.

src/Products/PluginIndexes/CompositeIndex/dtml/browseIndex.dtml
```diff
@@ -0,0 +1,64 @@
...
+              DateIndexes store dates packed into an integer, unpack
+              into year, month, day, hour and minute, no seconds and UTC.
+            --></dtml-comment>
+            <dtml-var "DateTime((_['sequence-key'] / 535680),
+                                (_['sequence-key'] / 44640 ) % 12,
+                                (_['sequence-key'] / 1440  ) % 31,
+                                (_['sequence-key'] / 60    ) % 24,
+                                (_['sequence-key']         ) % 60,
+                                0, 'UTC')">
+          <dtml-else>
```
hannosch: This got updated in DateIndex a while back:
```
<dtml-var "DateTime(((_['sequence-key'] - 44640) / 535680),
                    ((_['sequence-key'] - 1440) / 44640 ) % 12 or 12,
                    (_['sequence-key'] / 1440 ) % 31 or 31,
                    (_['sequence-key'] / 60 ) % 24,
                    (_['sequence-key'] ) % 60,
                    0, 'UTC')">
```

src/Products/PluginIndexes/CompositeIndex/help/CompositeIndex_searchResults.stx
```diff
@@ -0,0 +1,23 @@
...
+      'query' -- either a sequence of objects or a single value to be
+      passed as query to the index (mandatory)
+
+      'operator' -- specifies the combination of search results when
+      query is a sequence of values. (optional, default: 'or').
+       
+        Allowed values:
+
+         'and', 'or' 
+
```
hannosch: This file can go, together with the help entries in manage_options.

hannosch:
@andbag finally managed to take a look, all my review comments are in. Overall this looks really good. The ITransposeQuery interface is probably also really useful for solr/elasticsearch indices, which want to take over large parts of the query.

Extra bonus points, if you want to make this compatible with BooleanIndex :)

hannosch:
@andbag Looks like you have taken care of all my feedback, thx!

Do you want to continue working on this, or can I merge this?

andbag:
@hannosch Yes, you can merge my commits :). However, I cannot create a zope instance with current buildout.cfg for testing the ZMI.

[snip]
 While:
  Installing instance.
Error: There is a version conflict.
We already have: Zope2 2.13.24
but Products.PythonScripts 4.0 requires 'Zope2>=4.0a1'.

Do you have a working buildout.cfg for a Zope2 (4.0a1) instance?

hannosch:
@andbag I tested the ZMI part and added one final missing registerClass bit, to make the index show up in the "Add index" drop down lists.

To test, I added a new zopepy part to the buildout.cfg for ZCatalog:
```
parts = ... zopepy

[zopepy]
recipe = zc.recipe.egg
interpreter = zopepy
eggs = Products.ZCatalog
dependent-scripts = true
```
And than ran:
```
$ bin/buildout -N
$ bin/mkzopeinstance -d .
... <answer prompts>...
$ bin/runzope -C etc/zope.conf -X "debug-mode=on"
```

andbag:
@hannosch Thx, it works :)!

PR #18, opened 2016-09-13
Precision property for DateIndex and DateRangeIndex

andbag:
The effectiveness of date index caching can be strongly increased if the index would factor in the precision of the indexed time values. For example, the time precision of the indexed attributes "start", "end", "expires" and "effective" has a precision abut 5 minutes in Plone. There is no requirement to consider a higher precision for the indexing of these attributes. The default precision is currently one minute and is not configurable. Depending on the applied query values the TTL of the cache key can be shorter than necessary (e.g. for the "effectiveRange" index in Plone). The PR implements a configurable precision property incl. tests for the Date- and DateRangeIndex.

### DIFF
--- a/src/Products/PluginIndexes/DateIndex/DateIndex.py
+++ b/src/Products/PluginIndexes/DateIndex/DateIndex.py
@@ -1,4 +1,4 @@
-##############################################################################
+#############################################################################
 #
 # Copyright (c) 2002 Zope Foundation and Contributors.
 #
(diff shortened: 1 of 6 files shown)


hannosch:
@andbag thx, I'll look at this and your other PR soon, hopefully later this week.

hannosch:
For some definition of "soon", I've finally looked at at this. Rebased onto current master and slightly adjusted for Python 3 compatibility in #32. Also factored out the common parts of the `convertDateTime` method / test function into a `Products.PluginIndexes.util.datetime_to_minutes` function.

Closing this in favor of the new PR.

PR #31, opened 2017-11-30
Add new `getAllBrains` method to the ZCatalog.

hannosch:
The method returns a generator of brains for all cataloged objects. You can use this if you relied on `searchResults` returning all brains for empty queries before version 4.0a2.

Also add tests for the catalog's `__getitem__` and `clear` methods.

This is similar to the `catalog_get_all` method added to Plone's catalog_tool as part of plone/Products.CMFPlone@8520b2bc3ad6966e53673d92d164d06d05bf29e4.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Changelog
 4.0.2 (unreleased)
 ------------------
 
+- Add new `getAllBrains` method to the ZCatalog, returning a generator
+  of brains for all cataloged objects. You can use this if you relied
+  on `searchResults` returning all brains for empty queries before
+  version 4.0a2.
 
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,10 @@ Changelog
 4.0.2 (unreleased)
 ------------------
 
+- Add new `getAllBrains` method to the ZCatalog, returning a generator
```
icemac: Maybe this should become a 4.1 as this is a new feature.

icemac:
@thet Could you please review this PR, too.

pbauer:
Excellect addition. Now we can remove the ugly hacks we added in various places in Plone.

PR #34, opened 2018-03-02
Fix logging issue in KeywordIndex.

icemac:
The `debug` method of a logger does not accept an `error` keyword, see
https://docs.python.org/3.6/library/logging.html#logging.Logger.debug
resp.
https://docs.python.org/2/library/logging.html#logging.Logger.debug

This issue did not get noticed because the with the default log level in tests
this log message is discarded.


icemac:
Thank you all. I created a new 4.1 release including this fix.

PR #36, opened 2018-04-22
Fix TypeError

ale-rt:
This fixes an error when trying to get from an `OOBTree` a non existent value and the key has a type that cannot be compared with the existing keys

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changelog
 4.2 (unreleased)
 ----------------
 
-- Nothing changed yet.
+- Fix a TypeError on Python 3 when trying to lookup in an OOBTree
+  a value for a key that has an invalid type
 
 
 4.1 (2018-03-06)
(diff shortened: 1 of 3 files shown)


ale-rt:
This makes the behavior consistent with what you had on Python 2.7.
And in addition this fixes https://github.com/zopefoundation/Products.CMFUid/pull/7 ;)

hannosch:
I would add some kind of logging in this case. Something like:
```
LOG.error(
    '%s: query_index tried to look up key %s '
    'from index %s but key was of the wrong type.' %
        (self.__class__.__name__, repr(k), str(self.id)))
```

Since the code is in the ``query_index`` method, the error could happen as a result of bad user input and some upstream code not doing enough input validation. I'd not show a traceback to those users.

But as a developer the error log entry should still be helpful in finding the application level code problem. Some code is trying to query the index with a value of a wrong type after all. And likely this is bad old code, which so far failed silently or behaved unpredictably in the past.

ale-rt:
I added a log line, thanks for reviewing @davisagli and @hannosch

PR #44, opened 2018-10-02
Replace urllib.quote with six.moves.urllib.parse.quote

ale-rt:


CHANGES.rst
```diff
@@ -4,6 +4,8 @@ Changelog
 4.2 (unreleased)
 ----------------
 
+- Replace urllib.quote with six.moves.urllib.parse.quote
```
icemac: Does this fix the ZMI for this package? (If yes, this should be mentioned in the change log.)

PR #50, opened 2018-12-15
Fix "invalid escape sequence" warning.

jugmac00:
Beginning with Python 3.6 a DeprecationWarning is shown when an invalid
escape sequence is used. Converting the string into a raw string fixes
the problem.

cf
https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Changelog
 ----------------
 
 - Replace deprecated ``cgi.escape`` with ``html.escape`` for Python 3.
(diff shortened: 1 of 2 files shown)


dataflake:
> Is the "coverage decreased" failure a hard failure? I just added a single char to a test.

I don't know why it does that for tiny decreases. That should have a minimum threshold.

icemac:
> I don't know why it does that for tiny decreases. That should have a minimum threshold.

I set the threshold to 0.5 %.

Issue #55, opened 2019-03-07
`PluginIndexes.UnIndex._apply_index` sometimes forgets the intersection with `resultset`


d-maurer:
There are several weaknesses concerning the `resultset` parameter of `ILimitedResultIndex._apply_index`.

1. The documentation states `The additional resultset argument contains the resultset, as already calculated by ZCatalog's search method.` but does not specify what the index should do with it.

2. `Unindex._apply_index` sometimes intersects its primary result with `resultset` and sometimes forgets to do so -- as shown by the following example (using `Products.ZCatalog==4.2`, but the current master likely has the same problem):
```
>>> from Products.PluginIndexes.unindex import UnIndex
>>> ui = UnIndex("ui")
>>> docs={1:1, 2:2}
>>> for did, e in docs.items(): ui.insertForwardIndexEntry(e, did)
... 
>>> list(ui._apply_index(dict(ui=dict(query=(1,))))[0])
[1]
>>> list(ui._apply_index(dict(ui=dict(query=(1,))), IISet())[0])
[1]
>>> list(ui._apply_index(dict(ui=dict(query=(1,2))))[0])
[1, 2]
>>> list(ui._apply_index(dict(ui=dict(query=(1,2))), IISet())[0])
[]
```
i.e. if we search with a single value, then there is no intersection with `resultset` unlike a search for multiple values.

hannosch:
The resultset is an optional argument the index may use to perform its job. If the index logic can be more efficient by taking a pre-existing result into account, it may use the resultset. The API so far doesn't require the index to do anything with the resultset.

The intersection or weighted intersection of multiple index results happens inside the `catalog._search_index` method.

d-maurer:
Hanno Schlichting wrote at 2019-3-9 13:20 -0800:
>The resultset is an optional argument the index may use to perform its job. If the index logic can be more efficient by taking a pre-existing result into account, it may use the resultset. The API so far doesn't require the index to do anything with the resultset.

Thank you for the clarification.
However, the interface documentation should clearly state this.
Otherwise, index authors have no idea what they should/must do with
the argument `resultset`.


At other places, too, the "docstring" information is of poor
quality, e.g. `Products.PluginIndexes.unindex.UnIndex.query_index`:
```
    def query_index(self, record, resultset=None):
        """Search the index with the given IndexQuery object.

        If the query has a key which matches the 'id' of
        the index instance, one of a few things can happen:

          - if the value is a string, turn the value into
[...]

PR #57, opened 2019-03-08
refactor `PluginIndexes.unindex.UnIndex.query_index` fixing #55 and #56

### LINKED ISSUE #56: `PluginIndexes.Unindex.query_index` seems to be overly complex and should be refactored
* Its docstring is misleading
* There is duplicate code for cache and `not_param` handling
* Why is caching handled differently for operator "or" and not "or" (likely "and"). We could either cache in both cases a "setlist" or in both cases a set (with in the "and" case the intersection of the current "setlist"s).
Do we need this caching at all? How often do we have the same index query (with identical parameters) in the same request?

I propose the following new structure:
* we determine the cache key and obtain a cached result: either `None` or a sequence of sets
* if the cache has no result, we determine from *record* a sequence of keys to be looked up; for "and", the intermediate result is the sequence of looked up document sets, ordered by `len` (note that determining the size of an `IITreeSet` loads all buckets from the ZODB; if this requires real storage accesses, then this length determination can even be more expensive than performing the intersection directly); for "or", the result is the one element sequence of the "multiunion" of the looked up document sets. Cache the intermediate result
* Intersect with *resultset* and return the final result

d-maurer:
Fixes #56

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Changelog
 4.4 (unreleased)
 ----------------
 
+- Refactor ``Products.PluginIndexes.unindex.Unindex.query_index``
+  (`#56 <https://github.com/zopefoundation/Products.ZCatalog/issues/56>`_)
+  and fix
+  (`#55 <https://github.com/zopefoundation/Products.ZCatalog/issues/55>`_).
+
 - Specify supported Python versions using ``python_requires`` in setup.py
   (`Zope#481 <https://github.com/zopefoundation/Zope/issues/481>`_)
 
--- a/src/Products/PluginIndexes/FieldIndex/FieldIndex.py
+++ b/src/Products/PluginIndexes/FieldIndex/FieldIndex.py
@@ -21,6 +21,7 @@ class FieldIndex(UnIndex):
     """
     meta_type = 'FieldIndex'
     query_options = ('query', 'range', 'not')
+    potentially_multivalued = False  # optimization for exclude terms
 
     manage_options = (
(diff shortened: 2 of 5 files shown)


d-maurer:
@ale-rt 
> I think it should be `if not six.PY2 and isinstance(code, six.binary_type)`

I assume you wanted to comment on @icemac 's "widcode" changes (which is Python2/Python3 dependent). My PR is independent of Python 2/3 (and "six").

ale-rt:
You are right @d-maurer I was commenting on the merge commit :)
Did not follow a lot the development recently but it seems master already has a problem here:

https://github.com/zopefoundation/Products.ZCatalog/blob/1ec5a94919994b2867776c09ba27b56d39c9d335/src/Products/ZCTextIndex/WidCode.py#L88-L96

On Py2 decoding text (if possible) will return again text:
```
     if not isinstance(code, str): 
         code = code.decode('latin1') 
```

d-maurer:
@hannosch has convinced me that #55 is not a bug.

In my view, the code in `Products.PluginIndexes.unindex.UnIndex.query_index` is horrible (difficult to understand, difficult to maintain, difficult to read) but apparently, it does work.

A refactorisation would have helped me with the upcoming version of `Products.AdvancedQuery`. But, meanwihile I have found an (even more general) approach: `AdvancedQuery` will use the elementary indexing information but implement the filtering itself. Thus, I am no longer interested in the quality of the base implementations -- but still will keep you informed about bugs I detect (or think to have detected).

PR #60, opened 2019-03-10
#56: improve interface documentation, docstrings; add tests

d-maurer:


dataflake:
@icemac I need to take myself off the reviewer list, I never had anything to do with the indexing code and am not competent enough to do a review.

PR #67, opened 2019-04-08
Query with invalid paramter should raise an error

andbag:
Parameters can be used for a query that are not supported by the indexes. The user does not receive a feedback about the invalid used parameters. Results of a query could therefore misinterpreted. This behavior has already been identified https://github.com/zopefoundation/Products.ZCatalog/blob/dd30d4413eb1090f2a05bb4cfe438dc0a9690611/src/Products/PluginIndexes/FieldIndex/tests.py#L258-L259 but the strategy is simply to ignore the invalid parameter. I therefore suggest that instead of skipping invalid parameters, an error should be raised.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Changelog
 4.5 (unreleased)
 ----------------
 
-- Nothing changed yet.
+- Raise a ValueError if a query uses invalid index parameters. This prevents
+  the query from being changed without feedback to the user and delivering
+  implausible search results.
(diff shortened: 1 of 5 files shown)


andbag:
@icemac I have created an entry in the change log. I don't have any practice in code management for the community. Therefore I ask you to do the next steps as soon as the right time is reached.

icemac:
@andbag Thank you for fixing this issue. Should I cut a release now or are you planning other fixes which should go into the next release?

andbag:
@icemac Thank you for the merge, it would be nice if the PR #68 would also be included in the next release.

PR #70, opened 2019-04-12
Fix unindex convert issue #69

### LINKED ISSUE #69: UnIndex applies its `_convert` method on searching but not on indexing
There is another suboptimal thing with `UnIndex`:
it applies its `_convert` method on searching but not on indexing.

In general, it is adviceable to process the values for indexing
in the same way as those for searching - otherwise, there is
a risk that searches give surprising results.

The `_convert` of `UnIndex` is the identity. Thus, we do not see
a true error. However, if `UnIndex` would employ `_convert` homogenously
(both during indexing as well as during searching), then (e.g.)
`DateIndex` could  define its own `_convert`
and use everything else from `UnIndex/FieldIndex`.

_Originally posted by @d-maurer in https://github.com/zopefoundation/Products.ZCatalog/issues/64#issuecomment-477766804_

andbag:
The code is now consolidated for general usage of _convert to avoid unnecessary doubling of code.

Fixes #69.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,10 @@ Bug fixes
 - Fix rewriting of query to avoid wrong optimization of CompositeIndex.
   (`#59 <https://github.com/zopefoundation/Products.ZCatalog/issues/59>`_)
 
+- Consolidate code for general usage of ``UnIndex._convert`` method to
+  avoid unnecessary doubling of code.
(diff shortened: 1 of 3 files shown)


andbag:
@d-maurer Sorry for my copy & paste fault. The change log is fixed.

icemac:
@andbag You should be able to merge this PR.

PR #72, opened 2019-04-18
Code cleanup and completion of TopicIndex according to existing index methods and properties

### LINKED ISSUE #65: `TopicIndex`: uses `defaultOperator` in place of `useOperator`
... and thereby fails its `IQueryIndex` responsibilities.

andbag:
Fixes #65.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,7 +30,8 @@ Bug fixes
   when path of object changes
   (`#61 <https://github.com/zopefoundation/Products.ZCatalog/issues/61>`_)
 
-- Consequent use of ``QueryIndex`` capabilities 
+- Code cleanup and completion of TopicIndex according to existing index
+  methods and properties
  (`#65 <https://github.com/zopefoundation/Products.ZCatalog/issues/65>`_)
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,9 @@ Bug fixes
   when path of object changes
   (`#61 <https://github.com/zopefoundation/Products.ZCatalog/issues/61>`_)
 
+- Consequent use of ``QueryIndex`` capabilities 
+ (`#65 <https://github.com/zopefoundation/Products.ZCatalog/issues/65>`_)
+
 
 4.4 (2019-03-08)
 ----------------
--- a/src/Products/PluginIndexes/TopicIndex/FilteredSet.py
+++ b/src/Products/PluginIndexes/TopicIndex/FilteredSet.py
@@ -38,7 +38,7 @@ class FilteredSetBase(Persistent):
         self.ids = IITreeSet()
 
     def index_object(self, documentId, obj):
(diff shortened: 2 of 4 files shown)


andbag:
> Formerly, you have been exemplary with respect to tests (while I tend to avoid to test things that are obvious). This time, you have not provided a test.

This PR removes only redundant code. Operator tests for `and` and `or` already exist.
https://github.com/zopefoundation/Products.ZCatalog/blob/8658e8a9b0bf6a6e93ff041a57d02e49ac29fd8b/src/Products/PluginIndexes/TopicIndex/tests.py#L77-L89

andbag:
@d-maurer BTW, I fixed still some other poor implementations of methods defined by interfaces. Query plan initally considers indexes providing `IUniqueValueIndex` interface. However, TopincIndex does currently not have this property. Actually the implementation of the `uniqueValues` method is therefore not necessary. Should TopicIndex now be a UniqeValueIndex or not? If yes, then the `hasUniqueValuesFor` method is still missing.

d-maurer:
Andreas Gabriel wrote at 2019-4-19 14:57 -0700:
>@d-maurer BTW, I fixed still some other poor implementations of methods defined by interfaces.

Very good!

> Query plan initally considers indexes providing `IUniqueValueIndex` interface. However, TopincIndex does currently not have this property. Actually the implementation of the `uniqueValues` method is therefore not necessary.

I have not checked the details regarding the implemented interfaces.
I have seen the definition of `uniqueValuesFor` and concluded
that the index wants to implement the corresponding interface.

> Should TopicIndex now be a UniqeValueIndex or not?

I do not know.

Honestly, I have no clear understanding what "UniqueValueIndex"
should mean. If it should mean that each object is indexed under at most
one value, then "TopicIndex" is definitely no "UniqueValueIndex" -- because
an object can belong to more than one "filteredSet" (in
this case, "KeywordIndex" would also be no "UniqueValueIndex"
because an object can be indexed under several keywords).

The "uniqueValuesFor" is typically used to enumerate potential values.
There are 2 (quite separate) use cases:
the enumeration of potential search terms (e.g. in a `ZSearch`)
("TopicIndex" can enumerate those values)
and
the enumeration of potential document field values
(e.g. in an add or edit form) ("TopicIndex" cannot do this).


If the term "UniqueValueIndex" is well chosen, then some values related
to such an index should be unique. The most natural choice would be the
value(s) under which the index indexes a document.
I suppose however, that the term "UniqueValueIndex" was not optimally
chosen and a different term would have been more appropriate.
I suppose that "UniqueValueIndex" was introduced for the first
use case above (the enumeration of potential search terms)
and that something like "SearchTermEnumeratingIndex" would have
been a better term. Should this supposition be correct, then
"TopicIndex" could be defined as such an index.

andbag:
@d-maurer Thank you for the detailed explanation. I agree with your opinion and  ``TpicIndex`` implements now ``IUniqueValueIndex`` with my last commit.

Issue #35, opened 2018-03-21
Management of 'empty' index query


sgeulette:
Hi,
Since 3.0 version, it is possible to search on 'not': cool ! Empty index elements are not included in the results.
It would be nice to can also query on empty index in the catalog.
And then combine those 2 different queries if needed. 

What do you think about this improvement ?
Regards

andbag:
Could you please define an example for a better understanding?

sgeulette:
On a KeywordIndex ('tags'), storing by example tags ;-). 
If I want to find brains without tag, I cannot query on [] or None. 
A solution is that I manage an empty value ['no_tag'] in the index to query on this special value.
I think it would be better that the catalog manage itself this behavior and can query on empty value search.

vincentfretin:
I guess you'll need a TreeSet of rids for non indexed docs in the index, like it is done for example in hypatia FieldIndex (used by substanced)
https://github.com/Pylons/hypatia/blob/master/hypatia/field/__init__.py#L94
See my example here:
https://github.com/Pylons/hypatia/issues/9#issuecomment-374869769

andbag:
You can test this branch 

   https://github.com/zopefoundation/Products.ZCatalog/tree/not-parm-patch

Queries on [] should work. If that's what you needed, I'll create a PR.

icemac:
@sgeulette Where you able to try the branch @andbag suggested?

sgeulette:
I couldn't test it in Plone.

d-maurer:
> It would be nice to can also query on empty index in the catalog.

It would be difficult to implement this on the index level: the index by itself knows only about the documents it has indexed (itself), not about the documents known by the catalog. What you are calling for are the documents known by the catalog and not known by the index.

`Products.AdvancedQuery` allows you to formulate queries like this via `~ Indexed(index)` (i.e. search
for the documents not indexed by *index*). As `ZCatalog` does not have a general "not"; you would need a new index query parameter telling the index to look into the enclosing catalog and determine the set of its known objects. Making assumptions about the enclosing catalog (and how to determine its known objects) is not nice -- at least on the conceptual level.

andbag:
@d-maurer by the way, you identified a bug. The same argument applies to the 'pure not' operation. The index can currently only return documents that the index knows. Obects without a value (== None) belong to the result set of a 'pure not' operation.

d-maurer:
[...]

Issue #64, opened 2019-03-27
`UnIndex`: surprising behaviour if more than a single attribute is indexed


d-maurer:
`UnIndex` suggests that it can index a document under several attributes (--> `indexed_attrs`).
However, it does not document what this means.

The current implementation effectively uses at most one attribute for any given object -- the last one in `indexed_attrs` for which the object has a value.

I do not think that this behaviour is intensional:
* `index_object` iterates of `indexed_attrs` and updates `_index` and `_unindex` for each attribute for which the object has a value - with later attributes undoing the update of former ones; one could get the same effect without successive updates
* it would be more natural that the first attribute with a value succeeds (rather than the last one).

I suggest to clearly document what indexing under several attributes should mean and ensure that the implementation follows (naturally) the intension.

andbag:
I agree with you. I had already wondered about this behavior. If it should be changed, also the CompositeIndex must be adapted accordingly. https://github.com/zopefoundation/Products.ZCatalog/blob/dd30d4413eb1090f2a05bb4cfe438dc0a9690611/src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py#L250

d-maurer:
Andreas Gabriel wrote at 2019-3-28 13:34 -0700:
>I agree with you. I had already wondered about this behavior. If it should be changed, also the CompositeIndex must be adapted accordingly. https://github.com/zopefoundation/Products.ZCatalog/blob/dd30d4413eb1090f2a05bb4cfe438dc0a9690611/src/Products/PluginIndexes/CompositeIndex/CompositeIndex.py#L250

There is another suboptimal thing with `UnIndex`:
it applies its `_convert` method on searching but not on indexing.

In general, it is adviceable to process the values for indexing
in the same way as those for searching - otherwise, there is
a risk that searches give surprising results.

The `_convert` of `UnIndex` is the identity. Thus, we do not see
a true error. However, if `UnIndex` would employ `_convert` homogenously
(both during indexing as well as during searching), then (e.g.)
`DateIndex` could  define its own `_convert`
and use everything else from `UnIndex/FieldIndex`.

PR #74, opened 2019-05-08
Not indexed value support (MissingValue, EmptyValue)

andbag:
As discussed in issue #35  UnIndex can now support queries on MissingValue and EmptyValue. KeywordIndex implements currently the the new feature. I hope for active feedback.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,12 @@ Changelog
 5.1 (unreleased)
 ----------------
 
+- Nothing changed yet.
+
+
+5.0.1 (2019-06-17)
+------------------
+
 Bug fixes
 +++++++++
 


andbag:
@d-maurer thanks for the helpful comments. However, the old implementation of KeywordIndex does not keep empty values like '()' in _unindex.

```
>>> from Products.PluginIndexes.KeywordIndex.KeywordIndex import KeywordIndex
>>> index = KeywordIndex('foo')
>>> class Dummy: pass
>>> obj1 = Dummy()
>>> obj1.foo = ('a','b')
>>> index.index_object(1,obj1)
True
>>> tuple(index._unindex.keys())
(1,)
>>> obj2 = Dummy()
>>> obj2.foo = ()
>>> index.index_object(2,obj2)
True
>>> tuple(index._unindex.keys())# expect (1, 2, ) but
(1,)
```
If we want to use _unindex for 'pure not' queries, then `_unindex` should also collect the special values.

d-maurer:
Andreas Gabriel wrote at 2019-5-8 13:57 -0700:
>@d-maurer thanks for the helpful comments. However, the old implementation of KeywordIndex does not keep empty values like '()' in _unindex.

You are right.

>If we want to use _unindex for 'pure not' queries, then `_unindex` should also collect the special values.

You can do it -- but this is not a must:
the index knows what is in its `_unindex` and can compensate
as necessary. E.g. if it knows that `_unindex` does not contain
empty values, it could implement the "pure not" based on
"union(IISet(self._unindex), self._empty_...)".

d-maurer:
@andbag 
**ATTENTION** When we now index `MissingValue`, we may get in trouble with the strange behaviour described in #64, i.e. when the index indexes more than a single attribute. We need at least tests for this case.
The current behaviour is to iterate over all "indexed attributes" and give each attribute a chance to modify the index according to its value. This means effectively, that the last attributes with a value succeeds. When we change this and index even "missing value", then the last attribute will effectively win, whether it has or not. I am quite sure that this would be unexpected.

As I wrote in #64, I believe that the current behaviour is not what was really intended: it would be much more natural if the *first* rather than the *last* attribute with a value succeeds. Maybe, use the opportunity to document what it should mean the an index indexes several attributes, and maybe, we change the order in the process.

Whether or not we do something about the documentation for the "several indexed attribute" case (or even change the order), we must ensure that an attribute with a value has precedence over one without a value. We can distinguish both cases by checking the return value of `_index_object`.

I am unsure how to handle the case "empty value" (in contrast to "missing value"): should an attribute with a non empty value have precedence over one with an empty value? This question is relevant only for `KeywordIndex` like indexes. Should we say that "empty value" must be handled differently from "missing value", then potentially, we must change `_index_object` as well to differentiate both cases.

andbag:
Unfortunately, there are no tests yet that check the current behavior for multiple indexed attributes. I will submit a new PR for these tests so that we don't lose track if we change the current behavior.

andbag:
@d-maurer My observations show that the last attribute always wins, regardless of whether the value of last indexed attribute  is set or not. The same applies to the existence of the last indexed attribute. Following test is based on code of master branch: 
```
>>> from Products.PluginIndexes.KeywordIndex.tests import TestKeywordIndex
>>> test=TestKeywordIndex()
>>> index = test._makeOne('foobar', extra={'indexed_attrs': 'foo, bar'})
>>> class DummyContent(object):
...    def __init__(self, **kw):
...       for k in kw.keys():
...          setattr(self, k, kw.get(k))
... 
>>> index.index_object(0, DummyContent(foo=['NO']))
True
>>> index.index_object(1, DummyContent(foo=['NO'], bar=None))
True
>>> index.index_object(2, DummyContent(foo=['NO'], bar=''))
True
>>> tuple(index._index)
()
>>> tuple(index._unindex)
()
```
If the last attribute has a value, it is stored in the index.
```
>>> index.index_object(3, DummyContent(foo=['NO'], bar='YES'))
True
>>> tuple(index._index)
('YES',)
>>> tuple(index._unindex)
(3,)
```
In this regard, the option "indexed attributes" has no effect :(. That's why I don't think anyone's using the feature.

d-maurer:
Andreas Gabriel wrote at 2019-5-14 08:04 -0700:
>@d-maurer My observations show that the last attribute always wins, regardless of whether the value of last indexed attribute  is set or not.
> ...
Apparently, it is indeed worse than I had thought.

Obviously a bug. It makes no sense to let the name indicate
that an index can index several attributes while the index actually
indexes just the last one specified.

> ...
>In this regard, the option "indexed attributes" has no effect :(. That's why I don't think anyone's using the feature.

You might be right.

I see two options:

 * we raise an exception when more than a single attribute is indexed

 * we document the feature "indexed attributes" and ensure that
   the implementation follows the documentation -- at least
   for "our own" indexes.

andbag:
> You might be right. I see two options:
> * we raise an exception when more than a single attribute is indexed 
> * we document the feature "indexed attributes" and ensure that the implementation follows the documentation -- at least for "our own" indexes.

I prefer option one, because I don't want to implement features that nobody apparently requires. This feature can be much better implemented using a method that is executed by calling the single "indexed attribute". Which error fits best? TypeError or NotImplementedError?

d-maurer:
Andreas Gabriel wrote at 2019-5-15 08:18 -0700:
> ...
> Which error fits best? TypeError or NotImplementedError?

I would opt for `NotImplementedError` (with a good description
what is not implemented).

andbag:
Another curiosity is that unlike KeywordIndex FieldIndex indexes empty strings. And objects for which the last attribute does not exist are not indexed at all :|. FieldIndex supports multiple indexed attributes, but it seems to be buggy.
```
>>> from Products.PluginIndexes.FieldIndex.tests import FieldIndexTests
>>> test = FieldIndexTests()
>>> index = test._makeOne('foobar', extra={'indexed_attrs': 'foo, bar'})
>>> class DummyContent(object):
...    def __init__(self, **kw):
...       for k in kw.keys():
...          setattr(self, k, kw.get(k))
... 
>>> index.index_object(0, DummyContent(foo='NO'))
True
>>> index.index_object(1, DummyContent(foo='NO', bar=None))
True
>>> index.index_object(2, DummyContent(foo='NO', bar=''))
True
>>> tuple(index._index)
('', 'NO')
>>> tuple(index._unindex.items())
((1, 'NO'), (2, ''))

```

andbag:
@d-maurer I suggest that we generally do not support multiple indexed attributes (s. previous FieldIndex example).

d-maurer:
Andreas Gabriel wrote at 2019-5-16 08:44 -0700:
>@d-maurer I suggest that we generally do not support multiple indexed attributes (s. previous FieldIndex example).

I am fine with this.

andbag:
@d-maurer Does one have to guarantee that the private method `_index_object` exists? I would like to divide the method into two methods (e.g. `_getObjectDatum(obj)` and `_indexObjectDatum(documentId, datum)`). One that determines the final value to be indexed. And one that finally indexes the resulting value. This would eliminate the need to iterate over the fields (==`indexed_attrs`) in the `index_object` method.

d-maurer:
Andreas Gabriel wrote at 2019-5-17 04:35 -0700:
>@d-maurer Does one have to guarantee that the private method `_index_object` exists? I would like to divide the method into two methods (e.g. `_getObjectDatum(obj)` and `_indexObjectDatum(documentId, datum)`). One that determines the final value to be indexed. And one that finally indexes the resulting value. This would eliminate the need to iterate over the fields (==`indexed_attrs`) in the `index_object` method.

My personal feeling is that one should not eliminate methods (even
private ones)  or significantly change the logic using them
in minor versions -- especially not for base classes in a
package with `plugin` in their name.

`unindex.UnIndex` has a very good chance to be used as base class
for "foreign" indexes and such an index would likely say:
the inherited `indexObject` is good, I must just implement `_indexObject`.
Those indexes will break when `indexObject` no longer calls `_indexObject`.


I am not sure whether we should nevertheless make your change
(which I like -- the current handling obviously is buggy and
not well thought-out) in the current major version.
Porting `Products.AdvancedQuery` for the new catalog (and Zope4/Python3),
I recognized that the new catalog has many changes (incompatible from
the point of view of `AdvancedQuery`) such that it required a
new major version of `AdvancedQuery`. Thus, maybe, the current
major version of `Products.ZCatalog` is still young and putting
your change into this major version has a similar effect than starting
a new major version.

andbag:
> `KeywordIndex` line 76: `insertNotIndexed` should not have the argument `newKeywords` or should have a different name (such as `insertSpecialIndexEntry`).

Done.

> `KeywordIndex` line 75: you may need to remove not indexed entries if `oldkeywords` is `None`.

`_unindex` stores now the special values `missing` or `empty` for not indexed entries The problem is therefore solved.

> `KeywordIndex` lines 81, 88: you want `oldkeywords` to be an `OOSet` but you store it as a `list`.

Fixed. This inconsistency existed even before my adjustments. OOSet is only needed to determine the difference between old and new keywords. At the moment OOSet is persisted in `_unindex`, perhaps lists are the better choice. Are there advantages or disadvantages?

> `KeywordIndex` line 84: I expect an `insertNotIndex(...)` somewhere in this `else` block (as there was one in the `then` block). In addition, it might be necessary to remove a potential `MissingValue` from the special index.

Fixed.

> `KeywordIndex` line 120: not sure that an exception from the call should be silently swallowed. I suggest to at least log an entry.

Fixed, also for UnIndex.

> `KeywordIndex` lines 149, 73: inconsistent check for missing `_unindex` entry (`_marker` versus `None`).

Fixed.

> `KeywordIndex`: I suggest to rename `...NotIndexed` to `...SpeciallyIndexed` (or something similar) - as the document is indexed, just not in the "normal" way.

Done.
 
> `KeywordIndex`, line 154: maybe, you do not need this line (in case, that `_unindex` is set to [] for an empty value (as this is the case for the old `KeywordIndex`).

As discussed above,  `KeywordIndex` does not keep empty sets.

> The names `MissingValue` and `EmptyValue` are in "CamelCase" which indicates a class. Maybe, we should avoid "CamelCasing" for them to be more conformant with PEP8.

Done.

> `interfaces` line 288: `NotIndexedValue` should not be used by the application; maybe, we want to indicate this by prefixing the name with `_`.

Done. Renamed to `_SpecialIndexValue`

> `KeywordIndex` - "pure not": I have not seen a "pure not" special handling for `KeywordIndex`. However, if we include `MissingValue` in a "pure not" for "UnIndex", we must as well include `EmptyValue` for `KeywordIndex` -- this could be done in `UnIndex` (to avoid code duplication in `KeywordIndex`.

Done.

> `UnIndex`, line 565: the "pure not" should likely be implemented via `_unindex` rather than an enumeration of all keys (which may cause a huge `multiunion` to be executed). In addition: `KeywordIndex` might need that documents indexed under `EmptyValue` are included (by default) in a "pure not" result.

Done.

d-maurer:
Andreas Gabriel wrote at 2019-5-24 07:52 -0700:
>> `KeywordIndex` line 76: `insertNotIndexed` should not have the argument `newKeywords` or should have a different name (such as `insertSpecialIndexEntry`).
> ...
>Fixed. This inconsistency existed even before my adjustments. OOSet is only needed to determine the difference between old and new keywords. At the moment OOSet is persisted in `_unindex`, perhaps lists are the better choice. Are there advantages or disadvantages?

The pickle of a `list` is likely somewhat smaller than that
of an `OOSet` with the same elements -- but I do not think that
this is really important.


Potentially, there is a difference because an `OOSet` is a
persistent object while a `list` is not. As a consequence,
accessing the "value" for an object requires a separate ZODB load
for an `OOSet` while with a `list` this is not necessary
(the "value" is part of the "_unindex" bucket). Similar for modifications.

What is better depends on the access pattern and the size of the "value"s.
If the "value"s were typically large (I think this is not the
case) and the access to `_unindex` typically isolated (I think
this is the case), then an `OOSet` might have better typical
performance.

As access to the `_unindex` values is rather rare, I do not
think that this difference is of high importance.

andbag:
@d-maurer 
>     * `None` value is interpreted as "missing" (maybe, this should indeed be the case; 
>        it should then be documented, e.g. in the `IIndexingMissingValue` interface)

I have no particular preference. In the old implementation, `None` was not indexed either. In contrast, an empty string was indexed. If FieldIndex implements `IIndexingEmptyValue`, empty string is currently interpreted as special value `empty`.

Currently `_get_object_datum` returns `missing`, if the calculation results in `NoneType`, `AttributeError` or `TypeError` for `@implementer(IIndexingMissingValue)`. Otherwise, `empty` is returned when the truth value test returns `False` for `@implementer(IIndexingEmptyValue)`.

d-maurer:
Andreas Gabriel wrote at 2019-5-27 08:43 -0700:
>@d-maurer
>>     * `None` value is interpreted as "missing" (maybe, this should indeed be the case;
>>        it should then be documented, e.g. in the `IIndexingMissingValue` interface)
>
>I have no particular preference. In the old implementation, `None` was not indexed either. In contrast, an empty string was indexed. If FieldIndex implements `IIndexingEmptyValue`, empty string is currently interpreted as special value `empty`.

I recommend to treat the empty string as a "normal" (not a special)
value.

I recommend to treat `None` as a missing value (and document this).

>
>Currently `_get_object_datum` returns `missing`, if the calculation results in `NoneType`, `AttributeError` or `TypeError` for `@implementer(IIndexingMissingValue)`. Otherwise, `empty` is returned when the truth value test returns `False` for `@implementer(IIndexingEmptyValue)`.

That definitely is not good enough: `0` (= zero) is a perfectly
"normal" value for an index with integer values.

And, in my view, `''` is a "normal" value for an index with `str` values.


There is no need to introduce a special value, if you can search
for the value in the "normal" way.
We have introduced the special values, because they allow us to
search for things not searchable in the "normal" way.

andbag:
@d-maurer I've corrected the code and generalized it a bit. Before I improve the code, it would be nice if you could have a look at my changes. Especially the mapping of the special value mapping can now be configured and the purpose is documented in `interfaces.py`.
https://github.com/zopefoundation/Products.ZCatalog/blob/6d5fa3e8f94cd4a459ecee79be247285bc09dd29/src/Products/PluginIndexes/interfaces.py#L306-L308
The implementation looks like this
https://github.com/zopefoundation/Products.ZCatalog/blob/6d5fa3e8f94cd4a459ecee79be247285bc09dd29/src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py#L51-L54

d-maurer:
@andbag 
> @d-maurer ... Before I improve the code, it would be nice if you could have a look at my changes.

I find the idea good but have a few suggestions:
* `special_values` could get a better name and description. Let's start with "description":
      A `dict` mapping "exceptional" object values to a special value.
      When the index indexes an object, it derives an index specific value from the object, the so called "object value" (relative to this index). This process can result in an exception or produce a value which the index cannot index in the normal way.
      The attribute controls what should happen in such a case. It maps exceptions or values to one of the special values. If an exception not mapped occurs, it is reraised; if an object value is not mapped, it is indexed normally.
  A name like `map_to_special_value` would fit quite well with this description.
* `KeywordIndex` maps `()` to `empty`. However, a `KeywordIndex` related object value could be any sequence, not just `tuple`. It might be better to replace the `dict` by methods (e.g. `map_value` ("map *value* to a special value, if necessary") and `map_exception_to_special_value`).

src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py
```diff
@@ -58,59 +72,120 @@ def _index_object(self, documentId, obj, threshold=None, attr=''):
...
                 for kw in newKeywords:
-                    self.insertForwardIndexEntry(kw, documentId)
-                if newKeywords:
-                    self._unindex[documentId] = list(newKeywords)
-            except TypeError:
-                return 0
+                    try:
+                        self.insertForwardIndexEntry(kw, documentId)
+                        keys.append(kw)
+                    except TypeError:
```
d-maurer: I doubt that this exception handling is right: it does not index the object if one key cannot be indexed - and the problem is only reported via a log entry.
In my opinion, other alternatives would be better:
* log then ignore keys not indexable
* do not catch the exception (and let the whole operation fail)
* handle this `TypeError` in the same way as if it had occurred during the object value determination (e.g. map to `missing`).

In any case, the logic is at the wrong place. One would need similar logic for "update existing index info" and it should not be duplicated.

src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py
```diff
@@ -58,59 +72,120 @@ def _index_object(self, documentId, obj, threshold=None, attr=''):
...
+                                      kw=kw,
+                                      doc_id=documentId,
+                                      index=self.id))
+
+                        self.unindex_objectKeywords(documentId, keys)
+                        return 0
+
+                newKeywords = OOSet(newKeywords)
+
+            self._unindex[documentId] = newKeywords
```
d-maurer: The `_unindex` update could be done together with the "update existing index info" case.

src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py
```diff
@@ -58,59 +72,120 @@ def _index_object(self, documentId, obj, threshold=None, attr=''):
...
+        except KeyError:
+            pass
+
+        # normalize datum
+        if isinstance(newKeywords, basestring):
+            newKeywords = (newKeywords,)
         else:
             try:
                 # unique
                 newKeywords = set(newKeywords)
```
d-maurer: At another place, the keywords are collected in an `OOSet`. Using different set types increases the constraints placed on the usable types for keywords: `OOSet` requires orderability (as the `BTrees` package as a whole); `set` requires hashability. I recommend to use `OOSet` uniformly (and avoid the `tuple` recasting).

andbag:
>  * `KeywordIndex` maps `()` to `empty`. However, a `KeywordIndex` related
> object value could be any sequence, not just `tuple`. It might be better to
> replace the `dict` by methods (e.g. `map_value` ("map _value_ to a special value,
> if necessary") and `map_exception_to_special_value`).

@d-maurer I just can't imagine how I can implement such a method generically. In the end, the `_get_object_datum` method in combination with  `map_exception_to_special_value` already serves the purpose, doesn't it?

d-maurer:
Andreas Gabriel wrote at 2019-5-31 02:43 -0700:
>>  * `KeywordIndex` maps `()` to `empty`. However, a `KeywordIndex` related
>> object value could be any sequence, not just `tuple`. It might be better to
>> replace the `dict` by methods (e.g. `map_value` ("map _value_ to a special value,
>> if necessary") and `map_exception_to_special_value`).
>
>@d-maurer I just can't imagine how I can implement such a method generically. In the end, the `_get_object_datum` method in combination with  `map_exception_to_special_value` already serves the purpose, doesn't it?

@andbag Likely, you cannot have a fully generic implementation. Just a
sensible default which a derived class can override if the default
does not fit.

All is a question of class architecture.
In your architecture, we have an `_index_object` and a `_get_object_datum`
(which you name `_get_object_keywords` for `KeywordIndex`).
`_index_object` essentially gets an attribute name. Its task is to determine
the object's value from this attribute and then updates the index accordingly.
It uses `_get_object_datum` for the first subtask.
The task of `_get_object_datum` is to produce a value the index can
handle, either a "normal" or a "special" value.
Currently, you define a local function `_getSpecialValueFor`
inside `_get_object_{datum,keywords}` (you definitely want to put this
on the index level) which "generically" recognizes cases which require
the mapping to a special value (using the dict `special_values`).

I suggest a slightly different architecture. It has methods
`_index_object`, `_get_object_datum` (which likely should get official
rather than private)  and `map_value` and
a tuple attribute `exceptions_treated_as_missing`.
In this architecture, `_get_object_datum` could look like:
```
   exceptions_treated_as_missing = AttributeError, TypeError,
   def _get_object_datum(self, obj, attr):
       try:
           datum = getattr(obj, attr)
	   if safe_callable(datum):
	       datum = datum()
	   return self.map_value(datum)
       except self.exceptions_treated_as_missing:
	   return missing
```
The default implementation for `map_value` could be:
```
   def map_value(self, value):
       return missing if value is None else value
```
`KeywordIndex` could override `map_value` as
```
   def map_value(self, value):
       value = super(KeywordIndex,self).map_value(value)
       if value is not missing:
           ... handle atomic value ...
	   # at this place, *value* is expected to be a sequence
	   value = empty if not value else OOSet(value)
       return value
```

With Python 3, a new limitation arises: a `BTree` can no longer
have keys of different types (exception `None`).
`TypeError`'s are raised when the condition is violated.
The implementation draft above would result in a `missing`
when an object has keywords of different type.
Maybe, we do not want this behaviour.

src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py
andbag: In the original version there was a bug that could lead to inconsistencies in the index. Also, the problem was not logged. In order not to have to abolish the old behavior completely, I would prefer the first variant. Consequently, `_unindex` is only allowed to store indexable keywords.

src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py
andbag: Okay

src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py
andbag: Okay

icemac:
What a pity that Python 3.8 segfaults when starting the test. I cleaned the caches and tried to restart the Python 3.8 job.

icemac:
Cool, cleaning the TravisCI cache seems to do the trick.

src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py
andbag: **Correction**: Since the type OOSet is forced for keywords in the meantime, a TypeError can also be raised under python3 e.g. in the method `map_value`. For consistency reasons, TypeError is now always handled in the same way when determining the attribute value.

src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py
```diff
@@ -57,63 +64,113 @@ def _index_object(self, documentId, obj, threshold=None, attr=''):
...
+                oldSet = oldKeywords
+
+            if self.providesSpecialIndex(newKeywords):
+                self.insertSpecialIndexEntry(newKeywords, documentId)
+                newSet = OOSet()
+            else:
+                newSet = newKeywords = OOSet(newKeywords)
+
+            try:
+                fdiff = difference(oldSet, newSet)
```
d-maurer: This will fail if the keywords change type - and will let your index in a strange state. Assume "oldSet" to be `OOSet(['a', 'b']), "newSet" to be `OOSet([1, 2])`. Then under Python3, the `difference` will result in a `TypeError` which lets your document remain indexed under `oldSet` and gets newly indexed under `missing`.

Under Python 3, all indexed values must have a common type. Changing the keyword type will therefore not work (apart from constructed cases). Therefore, you should let an exception from the `difference` calls propagate (maybe log and provide a more specific error message) and not turn it into `missing`.

src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py
```diff
@@ -57,63 +64,113 @@ def _index_object(self, documentId, obj, threshold=None, attr=''):
...
+    def index_objectKeywords(self, documentId, keywords):
+        """ carefully index keywords of object with integer id 'documentId'
+        """
+
+        for kw in keywords:
+            self.insertForwardIndexEntry(kw, documentId)
 
     def unindex_objectKeywords(self, documentId, keywords):
-        """ carefully unindex the object with integer id 'documentId'"""
+        """ carefully unindex keywords of object with integer id 'documentId'
```
d-maurer: You are not "carefull" here (drop the word).

src/Products/PluginIndexes/interfaces.py
```diff
@@ -283,3 +283,63 @@ def make_query(query):
...
+    def __lt__(self, other):
+        return self.name < other.name
+
+
+class _IIndexingSpecialValue(Interface):
+    """Abstract marker interface to mark indexes which support a
+    special value query term."""
+
+    def map_value(value):
+        """ Map value, which is typically not generically indexable,
```
d-maurer: The "is typically not" is wrong.

I recommend:
```
def map_value(value):
  """Map (original) value to the value that should get indexed.

  The (original) value obtained from the object might not be indexable in the normal way.
  `map_value` gives you the chance to map it to a different, usually a special value in this case.
  """
```

src/Products/PluginIndexes/KeywordIndex/KeywordIndex.py
andbag: @d-maurer I'm beginning to wonder if it wouldn't be more sensible to escalate TypeError when a value in the keyword list is incompatible with the already indexed values. Otherwise the new values would have to be pre-validated before being indexed.

d-maurer:
Andreas Gabriel wrote at 2019-6-26 00:34 -0700:
>andbag commented on this pull request.
>
>
>
>>                  for kw in newKeywords:
>-                    self.insertForwardIndexEntry(kw, documentId)
>-                if newKeywords:
>-                    self._unindex[documentId] = list(newKeywords)
>-            except TypeError:
>-                return 0
>+                    try:
>+                        self.insertForwardIndexEntry(kw, documentId)
>+                        keys.append(kw)
>+                    except TypeError:
>
>@d-maurer I'm beginning to wonder if it wouldn't be more sensible to escalate TypeError when a value in the keyword list is incompatible with the already indexed values. Otherwise the new values would have to be pre-validated before being indexed.

I am with you in this regard.

PR #75, opened 2019-05-10
fixes #62 Migrated manage_catalogIndexes from dtml to zpt

### LINKED ISSUE #62: Sorting in /portal_catalog/manage_catalogIndexes is broken
The current view has no sorting applied for the indexes.
This was better in older versions where first of all it was sorted by defaul on the name of the indexes and second one could sort on every column.

dhavlik:
Restored sorting mechanism (adapted from and inspired by default folder listing)

fixes #62


dhavlik:
Unfortunately, this introduces an incompatibility with ZCTextIndex: If you have a ZCTextIndex in our indexes, you will get an Unauthorized, when accessing meta_type attribute and indexSize method from the template. While these objects got acquisition wrapped implicitly in dtml, it worked - but now they are not, resulting in this problem.

CHANGES.rst
```diff
@@ -4,7 +4,12 @@ Changelog
 5.1 (unreleased)
 ----------------
 
-- Nothing changed yet.
+Bug fixes
++++++++++
+
+- Fix sorting of index overview table in ZMI. Migrated the template from 
+  to zpt.
+  (`#62 https://github.com/zopefoundation/Products.ZCatalog/issues/62`)
```
icemac:
```suggestion
  (`#62 <https://github.com/zopefoundation/Products.ZCatalog/issues/62>`_)
```

This makes the code a link.

PR #89, opened 2019-12-05
Do not break when indexing multiple indexes

icemac:
When trying to reindex multiple indexes while having a progress handler it was breaking before with:

```
...
  File ".../Products.ZCatalog-5.0.1-py3.7.egg/Products/ZCatalog/ZCatalog.py", line 420, in reindexIndex
    pghandler.init('reindexing %s' % name, len(paths))
TypeError: not all arguments converted during string formatting
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changelog
 5.1 (unreleased)
 ----------------
 
-- Nothing changed yet.
+- Fix exception happaning when reindexing multiple indexes while having a
+  progress handler.
+  (`#89 <https://github.com/zopefoundation/Products.ZCatalog/pull/89>`_)
 
 
 5.0.1 (2019-06-17)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog
 5.1 (unreleased)
 ----------------
 
-- Fix exception happaning when reindexing multiple indexes while having a
+- Fix exception happening when reindexing multiple indexes while having a
   progress handler.
   (`#89 <https://github.com/zopefoundation/Products.ZCatalog/pull/89>`_)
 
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -4,7 +4,9 @@ Changelog
 5.1 (unreleased)
 ----------------
 
-- Nothing changed yet.
+- Fix exception happaning when reindexing multiple indexes while having a
+  progress handler.
```
jugmac00: Typo: happaning -> happening

src/Products/ZCatalog/ZCatalog.py
```diff
@@ -410,14 +410,12 @@ def manage_clearIndex(self, ids=None,
 
     @security.protected(manage_zcatalog_entries)
     def reindexIndex(self, name, REQUEST, pghandler=None):
```
jugmac00: I'd like to see a docstring added, as it is not entirely clear what name is, or should be. Obviously it can be a string or something else.

d-maurer:
Jürgen Gmach wrote at 2019-12-5 23:24 -0800:
>jugmac00 commented on this pull request.
>
>
>
>> @@ -410,14 +410,12 @@ def manage_clearIndex(self, ids=None,
>
>     @security.protected(manage_zcatalog_entries)
>     def reindexIndex(self, name, REQUEST, pghandler=None):
>
>I'd like to see a docstring added, as it is not entirely clear what name is, or should be. Obviously it can be a string or something else.

Note that a "docstring" has as side effect that the object
becames "Web publishable" (i.e. can be directly called via the Web).
Reindexing can take a long time; therefore, it might be interesting
to quite tightly control its use.

jugmac00:
@d-maurer Thank you for the hint. I know why I am a big fan of "explicit is better than implicit" :-)

src/Products/ZCatalog/ZCatalog.py
```diff
@@ -410,14 +410,14 @@ def manage_clearIndex(self, ids=None,
...
-        if isinstance(name, str):
-            name = (name, )
-
+        # This method does the actual reindexing of indexes.
+        # `name` can be the name of an index of a list of names.
```
jugmac00: Thank you @icemac ! This makes the codebase more accessible!

P.S.:
One more typo:  `of` -> `or`

icemac: Fixed on master.

PR #93, opened 2020-02-11
Fix DeprecationWarning: InstanceDict is deprecated.

jensens:


icemac:
This has been fixed in parallel on `master` by @dataflake and it is even already released.

Issue #101, opened 2020-05-06
"Catalog" and "Find Objects" tabs in the ZMI


georgpfolz:
At first I thought "Find Objects" does not work, as it returned "There are no objects in the Catalog.". The same is shown in the "Catalog" tab where the catalogued objects used to be displayed in former versions.

I then created a FieldIndex before running another "Find Objects" procedure: although no objects were shown, I could see some data in the "distinct values" column in the "Index" tab.

So the problem seems only to be that the objects are not shown in the ZMI (they can be accessed normally from a script).

d-maurer:
Georg Pfolz wrote at 2020-5-5 23:50 -0700:
> ...
>So the problem seems only to be that the objects are not shown in the ZMI (they can be accessed normally from a script).

This bug is caused by a recent change: formerly, an (effectively) empty
search returned all hits; now, it returns 0 hits.
`dtml/catalogView.dtml` (i.e. the implementation of the `Catalog` tab)
is not yet updated to this change.

It is possible that this bug shows itself only when your catalog
has no index named `path`.

georgpfolz:
> It is possible that this bug shows itself only when your catalog
> has no index named `path`.

Great, thank you! That works for me :)

icemac:
Closing this issue as there is #102 as follow-up.

PR #104, opened 2020-05-11
update `catalogView.dtml` to changed behavior for "empty searches"

### LINKED ISSUE #102: "dtml/catalogView.dtml" needs update to changed behavior for "empty" searches
This originally has been #101 .

`searchResults` has recently been changed regarding "empty" searches: formerly, an "empty" search returned all catalogued objects, now, it returns 0 results.

The above mentioned dtml file must be updated accordingly. Otherwise, the `Catalog` tab reports an empty catalog whenever the catalog lacks an index named `path`.

d-maurer:
This fixes #102 (and #101).

It also interprets a missing or empty `filterpath` as "no restriction" (instead of `"/"`). This allows a potentially available `path` index to index generalized paths (not necessary starting with `"/"`).

### DIFF
--- a/src/Products/ZCatalog/dtml/catalogView.dtml
+++ b/src/Products/ZCatalog/dtml/catalogView.dtml
@@ -4,7 +4,7 @@
 <main class="container-fluid">
 
     <dtml-let filterpath="REQUEST.get('filterpath', '')"
-              results="searchResults(path=filterpath) if filterpath else list(getAllBrains())">
+              results="searchResults(path=filterpath) if filterpath else searchAll()">
 <dtml-if results>
 
 <h1 class="form-label section-bar">Path filter</h1>
--- a/src/Products/ZCatalog/dtml/catalogView.dtml
+++ b/src/Products/ZCatalog/dtml/catalogView.dtml
@@ -3,8 +3,8 @@
 
 <main class="container-fluid">
 
-    <dtml-let filterpath="REQUEST.get('filterpath', '/')"
-              results="searchResults(path=filterpath)">
+    <dtml-let filterpath="REQUEST.get('filterpath', '')"
+              results="searchResults(path=filterpath) if filterpath else list(getAllBrains())">
 <dtml-if results>
 
 <h1 class="form-label section-bar">Path filter</h1>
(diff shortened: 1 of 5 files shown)


src/Products/ZCatalog/dtml/catalogView.dtml
```diff
@@ -3,8 +3,8 @@
 
 <main class="container-fluid">
 
-    <dtml-let filterpath="REQUEST.get('filterpath', '/')"
-              results="searchResults(path=filterpath)">
+    <dtml-let filterpath="REQUEST.get('filterpath', '')"
+              results="searchResults(path=filterpath) if filterpath else list(getAllBrains())">
```
ale-rt, line 7: Isn't this a performance issue?

d-maurer:
Alessandro Pisa wrote at 2020-6-18 08:49 -0700:
>@ale-rt commented on this pull request.
>
>
>
>> @@ -3,8 +3,8 @@
>
> <main class="container-fluid">
>
>-    <dtml-let filterpath="REQUEST.get('filterpath', '/')"
>-              results="searchResults(path=filterpath)">
>+    <dtml-let filterpath="REQUEST.get('filterpath', '')"
>+              results="searchResults(path=filterpath) if filterpath else list(getAllBrains())">
>
>Isn't this a performance issue?

I have defined a new method `searchAll`.
It is similar to `getAllBrains` but returns a data type
like the other searches (`LazyMap`) instead of an interator.
I have replaced the expensive `list(getAllBrains())` with `searchAll()`.

src/Products/ZCatalog/tests/test_zcatalog.py
```diff
@@ -307,6 +314,16 @@ def testSearchNot(self):
...
     # _getProgressThreshold
 
+    def test_catalogView(self):
+        catalog = makerequest(self._catalog)
+        # hack `getPhysicalPath` to avoid problem with the catalog plan
+        catalog.getPhysicalPath = None
+        # provide `ZopeVersion`
+        catalog.ZopeVersion = lambda *arg, **kw: 4
+        vr = catalog.manage_catalogView()
+        self.assertTrue("There are no objects in the Catalog." not in vr,
```
icemac, line 324:
```suggestion
        self.assertNotIn("There are no objects in the Catalog.", vr,
```

Maybe an assertion what's actually in `vr` would be helpful as asserting something is not there is a relatively weak assertion. There are so many reasons something might not be there.

d-maurer: The test directly verifies the faulty behavior introduced by changing the way empty searches are handled ("all objects" -> "no objects"). It caused the message `There are no objects in the Catalog` even though this was false. The test checks that this faulty behavior is gone. Testing that the template actually produces the desired output requires much more machinery. I checked this aspect visually.

PR #106, opened 2020-05-19
Fix linting issues

jugmac00:
... which came with the new flake8 version.

I'd like some advice on what to do what to do when a parameter has to be changed, like the lowercase l in some places here.

The functions are at module level, so I consider them to be public API and I hesitate to rename their parameters in order to not break code which relies on them.

Should we create a deprecation warning? If so, please hint to some source code where this is done already.

Thanks for you input!


icemac:
I think the `noqa` makers are okay for now.

jugmac00:
Ready for review.

Issue #54, opened 2019-02-09
resolve BBB warnings for 5.0


icemac:
There are comments like this in the code:
`# BBB ZCatalog 5.0`

They should be resolved in a new version for Zope 5.0

dataflake:
You seem to say that a ZCatalog version 5 is somehow a prerequisite for Zope 5.0? I don't think it is.

icemac:
I do not see it as a prerequisite either, but when should we clean up these BBB comments? They seem to require backwards incompatible changes. I thought it would be a good fit for for the Zope 5 release.

dataflake:
My concern is that we're creating unnecessary obstacles to a Zope 5 release. It seems most items on that project board are not real prerequisites. They were just pushed there because they're unresolved. The mini-sprint is only a single day, unfortunately, and there's only so much we can do.

icemac:
You are right: we could handle this issue after the Zope 5 release. I'd like to keep it on the list for now. Let's see how much we can tackle on the next sprint.

sallner:
So let's introduce some deprecation warnings as a first step.

vernans:
Those already were deprecated for some time. I looked through the code maintained by zopefoundation, plone and collective. This seems to be a managable task. 
Zopefoundation has no occurrences at the moment. The tests for Zope and ZTK seem to be running.
In Products.LinguaPlone, Products.ExtendedPath and plone.app.multilingualindexes there is like one import each that needs to be corrected. Maybe I'll do that now. Also in the plone training for solr alternatives there is a reference. 
In collective I found 20-somewhat (many also using AT, not relevant, I didn't look through thoroughly after seeing that) occurrences. 
One occurrence is in plone.app.contenttypes, but it is for the ATCT migrations. @pbauer possibly this is relevant, but it doesn't seem so.

dataflake:
Solved by #110

PR #110, opened 2020-09-28
#54 Remove deprecated.

vernans:
ie Products.PluginIndexes.common and Products.ZCatalog.Lazy


dataflake (changes requested):
In https://github.com/zopefoundation/Products.ZCatalog/issues/54#issuecomment-699837318 Steffen says he will add deprecation warnings. There needs to be some time between adding those and a release that removes the code. This is too fast.

vernans:
@dataflake My comment on this issue wasn't ready yet. I added some information in the discussion. [#54 (comment)](https://github.com/zopefoundation/Products.ZCatalog/issues/54#issuecomment-699981551)

dataflake (commented):
I cannot make a judgment here because I have no idea what code still uses these modules.

dataflake (approved):

dataflake:
On second glance @vernans is right: The modules your PR removes had been marked with deprecation warnings for ZCatalog 5.0. The next release will be 5.2, so we're already past that deadline. I will merge this.

PR #111, opened 2020-10-07
Add support for Python 3.9.

icemac:


buildout.cfg
```diff
@@ -2,12 +2,13 @@
...
-parts = 
-    interpreter 
-    test 
+parts =
+    interpreter
+    test
 
 [versions]
 Products.ZCatalog =
+RestrictedPython = >= 5.1
```
dataflake, line 11: Remove this pin, you're already pinning in setup.py

dataflake (dismissed):
One tiny change

buildout.cfg
icemac: Without this pin the tests fail. It actually removes the constraint set in the `versions.cfg` file used via `extends` as it requires `5.0` which is not compatible with the requirement from `setup.py`.

We can remove it when we switch the versions used in `extends` to Zope 5.

Issue #81, opened 2019-08-12
sort-on multiple indexes is broken


volkerjaenisch:
When doing a sort-on on multiple indexes the 
_sort_iterate_resultset 
sort_algorithm has to be used (independent of the limit settings).

The limit calculation in Catalog.py (intorduced in commit 939c8c4091bbaba8ae283f6e65a22b79de5ddefc)

def _sort_limit_arguments(self, query, sort_index, reverse, limit)

does not respect the possibility that limit is set to None to choose _sort_iterate_result set.

While in the algorithm choice code the test on limit is None (uselessly) persists:
```
       # Choose one of the sort algorithms.
        if iterate_sort_index:
            sort_func = self._sort_iterate_index
        elif limit is None or (limit * 4 > rlen):
            sort_func = self._sort_iterate_resultset
        elif first_reverse:
            sort_func = self._sort_nbest
        else:
            sort_func = self._sort_nbest_reverse

```
The n_best/n_best_reverse do the limiting on the first index, only. And then the sorting over the additional indexes on the already limited result set which leads to incorrect results.

The fix to the issue is quite simple (just check if we have more than on sort_index)

```
        # Choose one of the sort algorithms.
        if iterate_sort_index:
            sort_func = self._sort_iterate_index
        elif limit is None or second_indexes or (limit * 4 > rlen):
            sort_func = self._sort_iterate_resultset
        elif first_reverse:
            sort_func = self._sort_nbest
        else:
            sort_func = self._sort_nbest_reverse
```

I also noticed that some Plone packages in particular plone.app.querystring do no longer even mention (and wrongly document in the code) the possibility to search/sort on more than a single index.

[https://github.com/plone/plone.app.querystring/issues/92](https://github.com/plone/plone.app.querystring/issues/92)

icemac:
Thank you for documenting the issue. Could you please create a pull request with your suggested solution?

volkerjaenisch:
@icemac  Thank you for taking this problem seriously. I also do.

IMHO the tests for zcatalog at least for TestCatalogSortBatch are complete nonsense. The bug I found is a simple problem - bugs happen. The real problem is that there **is** a testcase testing this particular usecase and returning true.

I spend some hours to understand the underlying test material. At first I did not trust my eyes so i dug deeper. Please correct me anybody if I tell stupid things here!

From debugging the tests I learned the following:
[...]

PR #120, opened 2021-03-26
Fix reversed sorting by multiple index + limit

### LINKED ISSUE #108: Wrong sorting when using sort_limit
Scenario: I have a folder full of documents and to make its listing I use the following catalog query:

```python
        query = {
            'member_id': self.context.id,
            'portal_type': 'freitag.article.article',
            'sort_on': 'effective',
            'sort_order': 'reverse',
        }
```

While trying to make things *faster* (side note: this issue has nothing to do with performance) I decided to add a new query parameter to it: `'sort_limit': self.b_start + self.b_size * 2`, but no matter what number I put on the `sort_limit` the elements are sorted differently :shrug: 

As I guess that is quite difficult to get help, where would be the best place to add a test to check for that, so I can try to get a reproducible test to check it? :thinking:

ale-rt:
Fix reversed sorting by multiple index by forcing the ``_sort_iterate_resultset`` sorting method when we have more than one sorting index

Fixes #108

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ Changelog
 
 - Fix case where multiple indexes with similar name seperated by ``_`` were interpreted as options.
   (`#78 <https://github.com/zopefoundation/Products.ZCatalog/issues/78>`_)
+- Fix reversed sorting by multiple index by forcing the ``_sort_iterate_resultset``
+  sorting method when we have more than one sorting index
+  (`#108 <https://github.com/zopefoundation/Products.ZCatalog/issues/108>`_)
 
 
 6.0 (2020-10-08)
--- a/src/Products/ZCatalog/Catalog.py
+++ b/src/Products/ZCatalog/Catalog.py
@@ -1000,7 +1000,7 @@ class Catalog(Persistent, Acquisition.Implicit, ExtensionClass.Base):
         # Choose one of the sort algorithms.
         if iterate_sort_index:
             sort_func = self._sort_iterate_index
-        elif limit is None or (limit * 4 > rlen):
+        elif limit is None or (limit * 4 > rlen) or sort_index_length > 1:
             sort_func = self._sort_iterate_resultset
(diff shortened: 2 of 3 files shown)


ale-rt:
This is probably a fix for #81 as well, even if I see that there is more advanced work happening in PR #82

ale-rt:
Without this patch the added test fails like:
```
AssertionError: Lists differ: [97, 76, 52, 39, 29, 26, 25, 24, 17, 0] != [99, 98, 97, 96, 95, 94, 93, 92, 91, 90]

First differing element 0:
97
99

- [97, 76, 52, 39, 29, 26, 25, 24, 17, 0]
+ [99, 98, 97, 96, 95, 94, 93, 92, 91, 90]
```

dataflake (approved):
I don't have any knowledge about the inner workings of ZCatalog, so I hope Dieter has better input. The unit test appears to show that your changes work, that's all I can go by.

d-maurer (approved):
The change is obviously safe - but it may disregard optimization potential: reverse sorting can be optimized in the same way as forward sorting and multi-index sorting can be optimized recursively.

ale-rt:
I would merge this because it fixes an issue observed by many people, even if it might cost some CPU cycles. 
I believe it is better to have something that works slow rather than something that (in some case) doesn't work but it is faster.

I see #82 looks promising performance wise, but it seems to me it requires still some work before it gets merged.

Should I backport this patch to the 5.x branch?

ale-rt:
Thanks for the reviews!

dataflake:
> Should I backport this patch to the 5.x branch?

Unless there's someone is specifically asking for it I wouldn't bother.

agitator:
@ale-rt backport for 5.x would be great, since we're using that 5.x for Plone 5.2.x

ale-rt:
@agitator yeah I know, anyway I think we can even switch to latest master, I do not see it so hard

PR #122, opened 2021-05-17
Update for zope-product

icemac:


icemac:
This PR is needed to get GHA green again after their problems.

tox.ini
```diff
@@ -18,6 +18,7 @@ skip_install = true
 deps =
     setuptools < 52
     zc.buildout
+    zope.testrunner
```
jugmac00, line 21: How did this work before?

tox.ini
```diff
@@ -46,7 +47,6 @@ deps =
     # Useful flake8 plugins that are Python and Plone specific:
     flake8-coding
     flake8-debugger
-    flake8-string-format
```
jugmac00, line 49: Is this intentionally left out?

jugmac00 (approved):
Just for your info - it is not necessary to create a PR in order to re-run tests, you can just hit the button:

Screenshot from 2021-05-17 10-13-09

It does not hurt either :-)

LGTM with a couple of annotations

.github/workflows/tests.yml
```diff
@@ -4,14 +4,17 @@ name: tests
 
 on:
   push:
-    branches: [ master ]
   pull_request:
   schedule:
     - cron: '0 12 * * 0'  # run once a week on Sunday
+  # Allow to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
icemac, line 11: @jugmac00 This line was needed to be able to re-run the job via the web-UI. It was missing so I was not able to rerun CI without creating a new commit.

tox.ini
icemac: Probably because `zope.testrunner` is part of the `tests` extra.

tox.ini
icemac: Yes, it creates too much hassle to change all the strings so they match the wishes of this checker.

.github/workflows/tests.yml
icemac: @jugmac00 Interesting you found a hidden switch to re-run the tests without this line.

icemac:

jugmac00:
Thanks for creating this PR and taking the time to answer my questions :)

PR #124, opened 2021-10-28
Optimize plain "not" queries

jensens:
In Plone 6 we do queries returning a result except one item (the current context). A `not` query on the UID catalog is used to do this. In larger projects with several 10 or 100 thousands object cataloged, this slowed down queries to take several seconds. Without the `not` everything was fast.

After some profiling we found the time is wasted in this loop https://github.com/zopefoundation/Products.ZCatalog/blob/7f614a6394e236beae6ba69053bfb8c04d3f073a/src/Products/PluginIndexes/unindex.py#L575-L664 primary while factoring IISets here https://github.com/zopefoundation/Products.ZCatalog/blob/master/src/Products/PluginIndexes/unindex.py#L608

Whats happens?
The `not` query is written to return all values except the ones matching. Thus, for a single UID, this creates an result containing all index keys, except one. This is very expensive.

What I have done:
If the code detects a simple `not` query (without any operators) excluding only one or more keys, I shortcut the whole `index_query` and return the previous result w/o the current catalog id.

Does it help?
Yes. We bench-marked our query on a customer database with ~367000 catalogued objects. It cut down the whole requests time from ~1200ms to ~80ms (with 40ms of it been not query time in both).

cc @tisto @mauritsvanrees @cekk

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,7 +3,7 @@ Changelog
 
 6.1 (unreleased)
 ----------------
-
+- Improve performance of simple ``not`` queries on large catalogs.
 - Fix case where multiple indexes with similar name seperated by ``_`` were interpreted as options.
   (`#78 <https://github.com/zopefoundation/Products.ZCatalog/issues/78>`_)
 - Fix reversed sorting by multiple index by forcing the ``_sort_iterate_resultset``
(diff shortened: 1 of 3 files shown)


jensens:
this branch includes now #125

ale-rt (approved):
LGTM

ale-rt:
We see that it is only possible to squash and merge the commits:
image
Is that intended?

icemac:
@ale-rt Yes this is intended to keep the commit history straight.

PR #126, opened 2021-11-09
Add support for Python 3.10

icemac:


dataflake (approved):
You're missing a change log entry for Python 3.10

icemac:
@dataflake Good catch, fixed it in the next commit.

PR #128, opened 2022-01-21
Refactored ZMI templates for Zope4+

drfho:
Hi @dataflake ,
I fitted ZCatalog-ZMI pages, which did not contain all the necessary CSS classes needed for a  nice Zope look. Following tabs are refreshed now:

- Catalog
- Query Report 
- Query Plan

_Screen Image: Left: former ZMI, right: refreshed ZMI_
ZCatalog_ZMI

A small side effect to be discussed:
Historically the Zope default CSS _zmi.base.css_ compensates some super-bad layout effects of the non-bootrappyfied ZCatalog ZMI. These 3 elements are not needed anymore for Zope: https://github.com/zopefoundation/Zope/blob/b72715cf01cce1a49b59198047113f29f89037c5/src/zmi/styles/resources/zmi_base.css#L891-L905
But many installations may not use the latest ZCatalog, so I added a CSS snippet into 
ZCatalog/dtml/catalogView.dtml Line134+ for neutralizing the upper Zope CSS dealing with the former non-conformance to a strict styling.  Please tell me, if you prefer removing it both from Zope an ZCatalog.

### DIFF
--- a/src/Products/ZCatalog/dtml/catalogObjectInformation.dtml
+++ b/src/Products/ZCatalog/dtml/catalogObjectInformation.dtml
@@ -1,103 +1,67 @@
-<dtml-var manage_page_header>
+<h3 class="mt-2">
+	Catalog record at 
+	<code>
+		<dtml-var expr="getpath(_.int(rid))" html_quote>
+	</code>
+</h3>
+<p>
+	The goal of this page is to provide basic debugging information
+	about what is in the Catalog on a specific object. Listed below is
+	all the information that the Catalog currently contains for the
+	above specified object. This information should match what is
+	currently in the instance of that object.
+</p>
 
-<main class="container-fluid">
-
-<table width="100%" borders="0" cellspacing="2" cellpadding="0">
-<tr bgcolor="#000000">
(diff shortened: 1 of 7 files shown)


drfho:
Instead of showing cataloged object info as a new window (and loosing the UI context) it is shown now as a modal popup:

ZCatalog_ZMI2

dataflake:
All these ideas are good (and I'd suggest leaving the superfluous CSS instead of breaking some installations), but I'd be so much happier if you could teach your HTML editor to not use tab characters for indentation. Tabs are forbidden in our Python code and I wish we could get rid of them everywhere else as well.

drfho:
Many thanks for the quick feedback 
Please excuse using tabs (to me tabs just were easier to handle). Actually I forget/didnt know, that tabs are forbidden in the zope project.  I will change it, when touching any code next time:
https://github.com/zopefoundation/Products.ZCatalog/pull/128/commits/0e95efed0693259f6cd344509220e807e35743e0

dataflake (approved):
Great work, thank you!

PR #135, opened 2022-04-13
Update 5.x to current meta/config.

icemac:
(This seems the only place where we have meta/config on a maintenance branch.)

We cannot support Python 3.10 because it requires a version of the `multipart` library which is not allowed by the Zope 4.x version pins used.

I had to tune the result of `meta/config` to match here.


dataflake (approved):

icemac:

Issue #140, opened 2022-11-16
Bad test (or code): `test_getCatalogPlan_partial`


dataflake:
The unit test `test_getCatalogPlan_partial` routinely fails on the macOS GitHub Actions runners when run as part of the Zope `alltests` test script, which tests Zope in conjunction with other important Zope packages, see traceback and GHA runner log output URL below. This test never fails for me on my local macOS setup. This leads me to believe that either the code being tested or the test itself can show race conditions in certain environments.

```
Failure in test test_getCatalogPlan_partial (Products.ZCatalog.tests.test_plan.TestCatalogPlan)
Traceback (most recent call last):
  File "/Users/runner/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/Users/runner/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/unittest/case.py", line 605, in run
    testMethod()
  File "/Users/runner/work/Zope/Zope/.tox/py36/eggs/Products.ZCatalog-6.3-py3.6.egg/Products/ZCatalog/tests/test_plan.py", line 313, in test_getCatalogPlan_partial
    cat.getCatalogPlan(query2).plan(), ["numbers", "num", "date"]
  File "/Users/runner/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/unittest/case.py", line 829, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Users/runner/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/unittest/case.py", line 1028, in assertListEqual
    self.assertSequenceEqual(list1, list2, msg, seq_type=list)
  File "/Users/runner/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/unittest/case.py", line 1010, in assertSequenceEqual
    self.fail(msg)
  File "/Users/runner/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/unittest/case.py", line 670, in fail
    raise self.failureException(msg)
AssertionError: Lists differ: ['numbers', 'date', 'num'] != ['numbers', 'num', 'date']

First differing element 1:
'date'
'num'

- ['numbers', 'date', 'num']
+ ['numbers', 'num', 'date']
```

Example at https://github.com/zopefoundation/Zope/actions/runs/3479685094/jobs/5818540212

dataflake:
Assigning @mamico - he added the test code.

[...]

PR #141, opened 2022-12-12
Make test_getCatalogPlan_partial test more stable.

mauritsvanrees:
On Mac you would sometimes get a different query plan.
Fixes https://github.com/zopefoundation/Products.ZCatalog/issues/140.
(Or at least I hope that this fixes the issue.)

The first commit updates to the latest meta config.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-        - ubuntu
+        - ["ubuntu", "ubuntu-20.04"]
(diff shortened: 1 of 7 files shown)


mauritsvanrees:
All tests pass, but we are not running any MacOS tests... Let me fix that in the config.

dataflake (approved):

mauritsvanrees:
Thanks for the review.
Can someone make a release? Then I can add this version in Zope, and we can see if this really helps.

dataflake:
I'll do it right now

PR #143, opened 2022-12-22
Drop support for Python 2.7, 3.5, 3.6.

### LINKED ISSUE #112: Drop Python 2 support and test against Zope 5
Version 6.0 is not intended to be used together with Zope 4 as it removes DeprecationWarnings which have to be resolved to smoothly migrate to Zope 5.

Currently we are testing against the versions pinned for Zope 4 because we still support Python 2.

To clean this up Python 2 support should be dropped (I see no problem here as ZCatalog is not a low level package but cannot be used outside Zope) and we should extend the versions pinned on `master` in `buildout.cfg`.

icemac:
Fixes #112.


src/Products/PluginIndexes/DateIndex/tests.py
```diff
@@ -184,9 +184,9 @@ def _convert(self, dt, precision=1):
         from Products.PluginIndexes.DateIndex.DateIndex import Local
         if isinstance(dt, (float, int)):
             yr, mo, dy, hr, mn = gmtime(dt)[:5]
-        elif type(dt) is date:
+        elif isinstance(dt, date):
```
mamico, line 187: The tests are broken because of this change: a `datetime` object is also a `date` by inheritance. So `type(dt) is date` is different than `isinstance(dt, date)` when talking about a `datetime`.

Anyway IMO tests could be easy fixed changing `if` order, as:

```
         from Products.PluginIndexes.DateIndex.DateIndex import Local
         if isinstance(dt, (float, int)):
             yr, mo, dy, hr, mn = gmtime(dt)[:5]
-        elif isinstance(dt, date):
-            yr, mo, dy, hr, mn = dt.timetuple()[:5]
         elif isinstance(dt, datetime):
             if dt.tzinfo is None:  # default behavior of index
                 dt = dt.replace(tzinfo=Local)
             yr, mo, dy, hr, mn = dt.utctimetuple()[:5]
+        elif isinstance(dt, date):
+            yr, mo, dy, hr, mn = dt.timetuple()[:5]
         else:
             yr, mo, dy, hr, mn = dt.toZone('UTC').parts()[:5]
         value = (((yr * 12 + mo) * 31 + dy) * 24 + hr) * 60 + mn

```

mamico (changes requested):

src/Products/PluginIndexes/DateIndex/tests.py
mamico: In the end here https://github.com/zopefoundation/Products.ZCatalog/blob/master/src/Products/PluginIndexes/DateIndex/DateIndex.py#L125 the order is as I suggested.

icemac: I applied the fix you suggested in https://github.com/zopefoundation/Products.ZCatalog/pull/143/commits/56d17f7363b7c6c7903af24936c24bc7661deb36

icemac:
Now we seem to have the same problem as in https://github.com/zopefoundation/Zope/pull/1083 for MacOS 3.9 and 3.10.

dataflake:
The GHA configuration needs to use `macos-11` instead of `macos-latest`. I have added this change to meta/config in https://github.com/zopefoundation/meta/pull/179/commits/35a1a7226e28806fd7cea018e68902fb0dc3206a. Do a git pull and re-run it against this repository.

icemac:
Done in https://github.com/zopefoundation/Products.ZCatalog/pull/143/commits/143e2d547276c34280b4f604d9f58a04da024f6e.

dataflake (approved):

icemac:

PR #152, opened 2024-05-25
Improve Handling of Empty Queries

kenara:
### Summary

This PR addresses the issue of empty queries in the ZCatalog returning no results without explicit feedback. It proposes raising a `ValueError` when an empty query is detected, providing clear feedback to developers.

### Background

In the current implementation of ZCatalog in Zope 4 and 5, an empty query returns no results without any indication to the user. This behavior can be confusing and complicates debugging.

### Justification

- **Improved User Experience**: Providing explicit feedback helps users understand why no results are returned, reducing confusion and aiding in debugging.
- **Consistency with Modern Systems**: Other indexing systems like Elasticsearch and Solr handle empty queries transparently, either by returning all documents or providing clear feedback.
- **Error Prevention**: Encouraging explicit query parameters reduces the risk of unintentional empty queries, which can improve the reliability of applications using ZCatalog.

### Changes

- Add a check in the `searchResults` method to raise a `ValueError` when an empty query is detected.
- Log a warning when an empty query is processed.
- Update documentation to reflect this new behavior, providing guidance on how to handle empty queries.

### Benefits

- **Improved Developer Experience**: Clear feedback helps developers understand and debug their queries more effectively.
- **Alignment with Modern Practices**: Makes ZCatalog's behavior consistent with other widely-used indexing systems.
- **Error Prevention**: Reduces the risk of unintentional empty queries, improving application reliability.

### Impact

- **Developers**: Clearer understanding of query failures, leading to faster debugging and development cycles.
- **Performance**: Minimal impact as the change involves simple conditional checks and logging.
- **Compatibility**: Maintains backward compatibility by raising errors only when queries are genuinely empty.

### DIFF
--- a/src/Products/ZCatalog/ZCatalog.py
+++ b/src/Products/ZCatalog/ZCatalog.py
@@ -612,6 +612,11 @@ class ZCatalog(Folder, Persistent, Implicit):
 
         Search terms can be passed as a query or as keyword arguments.
         """
+        # Check if the query is empty
+        if not kw:
+            LOG.warning("Empty query: No search parameters provided.")
+            raise ValueError("Empty query: No search parameters provided.") 
+            
         return self._catalog.searchResults(query, **kw)
 
     security.declareProtected(search_zcatalog, '__call__')


d-maurer:
kenara wrote at 2024-5-25 07:39 -0700:
>This PR addresses the issue of empty queries in the ZCatalog returning no results without explicit feedback. It proposes raising a `ValueError` when an empty query is detected, providing clear feedback to developers.

The catalog usually returns the intersection of the subquery results.
As a consequence, an empty query should return all possible
results (as `ZCatalog` did in its early days).

PR #153, opened 2024-06-18
Drop support for Python 3.7.

icemac:


icemac:
Additionally to the broken MacOS jobs the code for setting the branch protection rules seems to be wrong here.

dataflake:
Looks like this is the first `zope-product` project that has set `with-macos = true` other than Zope itself. Obviously, there's a bug. But is it really necessary to run macOS tests here?

dataflake:
P.S.: @tseaver had a good point recently that running a bunch of tests that aren't really important are a waste of energy. We're continuing to do that in many places because we're not paying for it.

.github/workflows/tests.yml
```diff
@@ -61,7 +56,11 @@ jobs:
         python -m pip install --upgrade pip
         pip install tox
     - name: Test
+      if: ${{ !startsWith(runner.os, 'Mac') }}
       run: tox -e ${{ matrix.config[1] }}
+    - name: Test (macOS)
+      if: ${{ startsWith(runner.os, 'Mac') }}
+      run: tox -e ${{ matrix.config[1] }}-universal2
```
davisagli, line 63: I think this causes the failures on macos since this environment isn't defined in tox.ini?

davisagli (commented):
Running the tests on macos was added in https://github.com/zopefoundation/Products.ZCatalog/pull/141 because the query plan was sometimes different there.

.github/workflows/tests.yml
dataflake: Yes it does, I know.

icemac:
Testing on MacOS was added in #141 because of the bug in #140. It seems to behave differently that's why we are testing it.

icemac:
Note: I do not plan to work on fixing this PR.

dataflake:
I will fix it.

dataflake (approved):

dataflake:
The fix for macOS is the same as already used for Zope itself: C-based packages must be pre-installed with pip because zc.buildout/setuptools messes it all up. This dependency pre-install might be moved from individual packages' meta configuration into the meta templates itself, if I encounter another package like that I'll handle it.

icemac:
Thank you for working on this PR. 😃

PR #156, opened 2024-12-02
Update Python version support.

icemac:
- **Add support for Python 3.13.**
- **Configuring for zope-product**
- **Update Python version support.**

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,33 @@
+# Generated from:
+# https://github.com/zopefoundation/meta/tree/master/config/zope-product
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches:
+    - master
+  # Allow to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+env:
+  FORCE_COLOR: 1
+
+jobs:
+  pre-commit:
+    name: linting
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: 3.x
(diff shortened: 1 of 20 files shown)


buildout.cfg
```diff
@@ -9,6 +9,7 @@ parts =
 [versions]
 Products.ZCatalog =
 RestrictedPython = >= 5.1
+zope.interface = >= 7.2
```
icemac, line 12: Without this line `pip` installs 7.2 but `buildout` wants to install 7.1.1 (as pinned in  https://zopefoundation.github.io/Zope/releases/master/versions.cfg). This fails on MacOS  for known reasons: buildout cannot handle the universal wheels.

It's a pitty that we have to have it here … maybe updating the master versions would be a better solution.

icemac:
coveralls is set to require at least 85.0 % coverage but do not allow to drop more than 0.5 %.
I think that happend because of a newer `coveralls` version. So lets hope this vanishes when merging to master.

buildout.cfg
dataflake: There's no reason you couldn't just update the pin in Zope and then update the gh-pages branch. I am sure it won't break anything.

icemac: Done in https://github.com/zopefoundation/Zope/pull/1241

dataflake: So I guess you can remove that pin now.

dataflake: I updated the Zope gh-pages branch right now so when the new pin is published everything should be OK

icemac: Thank you. I removed the pin now.

icemac:
@dataflake Could you please approve this PR?

dataflake (approved):

PR #158, opened 2025-05-17
Searches with "not" in a KeywordIndex do not return records that do not contain a value for the index

### LINKED ISSUE #148: Searches with "not" in a KeywordIndex do not return records that do not contain a value for the index
## BUG/PROBLEM REPORT / FEATURE REQUEST

Searches with "not" in a `KeywordIndex` do not return records that do not contain a value for the index. For example, Plone's index Subject.

See this issue: https://github.com/plone/Products.CMFPlone/issues/3895

### What I did:

1. Create a document with the Tag: "Bulletin"
2. Create a Python script in ZMI with the content:

```python
return str(context.portal_catalog(Subject={"not": ["Bulletin"]}))
```
3. Run the script multiple times.

### What I expect to happen:

The search must return all content that does not contain the "Bulletin" Subject. Including objects that do not have any Subject.

### What actually happened:

Content that does not have a Subject is not returned.

### What version of Python and Zope/Addons I am using:

Zope 5.9
Python: 3.11

mamico:
This is a try to fix #148.

Following the workaround implemented by @gogobd, the code here adds a marker value for objects that have the attribute to be indexed but without any value.

However, there is an open problem related to the type of value defined for the marker, since BTrees doesn't allow values of different types, the marker value must necessarily be of the same type as the values defined for the index (in the test and fix in KeywordIndex, strings were used, but if you modify the test with integers, you'll immediately see that the test fails again)

\cc @jensens @wesleybl


davisagli:
@jensens Maybe the set of all doc ids could be fetched from the catalog via Acquisition without changing the API. I haven't tried it though...

mamico:
> @jensens Maybe the set of all doc ids could be fetched from the catalog via Acquisition without changing the API. I haven't tried it though...

@davisagli I did here: https://github.com/zopefoundation/Products.ZCatalog/pull/159 (as simple as I could)

PR #159, opened 2025-06-08
Searches with "not" in a KeywordIndex do not return records that do not contain a value for the index (2nd)

mamico:
This is another try to fix https://github.com/zopefoundation/Products.ZCatalog/issues/148

See also https://github.com/zopefoundation/Products.ZCatalog/pull/158

### DIFF
--- a/src/Products/PluginIndexes/unindex.py
+++ b/src/Products/PluginIndexes/unindex.py
@@ -447,6 +447,16 @@ class UnIndex(SimpleItem):
 
         operator = record.operator
 
+        # If not is the first operator, we need to start with the full
+        # resultset from the catalog
+        if not_parm and resultset is None:
+            try:
+                resultset = IISet(self.aq_parent.uids.values())
+            except AttributeError:
+                # this is needed for tests, or where indexes are used outside
+                # of a catalog
+                pass
+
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,11 @@ Changelog
 7.2 (unreleased)
 ----------------
 
+- Fix Searches with "not" do not return records that do not contain
+  a value for the index.
```
davisagli, line 8:
```suggestion
- Fix results for "not" queries in UnIndex-based indexes.
   Depending on the query plan, the results previously could be missing records for which the index found no value.
```

CHANGES.rst
```diff
@@ -4,6 +4,11 @@ Changelog
 7.2 (unreleased)
 ----------------
 
+- Fix results for "not" queries in UnIndex-based indexes.
+   Depending on the query plan, the results previously could be missing records for which the index found no value.
```
davisagli, line 8:
```suggestion
  Depending on the query plan, the results previously could be missing records for which the index found no value.
```

davisagli:
@icemac Could you help make a new release please? I don't have access on PyPI.

PR #167, opened 2026-05-29
Fix SyntaxWarning: 'return' in a 'finally' block.

icemac:
Fixes #166


src/Products/ZCatalog/ZCatalog.py
```diff
@@ -965,9 +965,11 @@ def expr_match(ob, ed):
     r = 0
     try:
         r = e.eval(md)
+    except Exception:
+        pass
```
davisagli: It didn't swallow exceptions before. I don't think we should start doing that. (Also, with this, if there is an exception, then `r` will not be defined and there will be a NameError on line 972)

icemac: You are absolutely right, I gave it another try.

dataflake (approved):

